PR #3, opened 2024-11-07
Add support for SSE KMS on S3

cyberdelia:
This adds support for [Server Side Encryption via KMS on S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html), by adding a new option to S3 secret named `kms_key_id`, and inject the proper headers to use the provided key identifier when present.

```sql
CREATE SECRET encrypted (
    TYPE S3,
    PROVIDER CREDENTIAL_CHAIN,
    CHAIN 'config',
    REGION 'eu-west-1',
    KMS_KEY_ID 'arn:aws:kms:region:acct-id:key/key-id',
    SCOPE 's3://bucket/sub/path'
);
```

This was tested against S3 directly, there is no tests against minio, as this seems it requires another piece of software (KES).

This is the same PR as [duckdb#14475](https://github.com/duckdb/duckdb/pull/14475).


samansmink:
Looks good! I guess we will need to look into adding test infra for this in our test aws account

PR #6, opened 2024-12-11
Fix timeout to default to 30 seconds instead of 30000 seconds

Mytherin:
Found by @ywelsch

httplib splits the timeout into two components (sec and usec): 

```cpp
Server &set_write_timeout(time_t sec, time_t usec = 0);
```

We were passing the timeout into the seconds component:

```cpp
client->set_write_timeout(http_params.timeout);
```

With a somewhat high default value of 30000. This fixes it so the default is 30 seconds.


carlopi:
This depends on https://github.com/duckdb/duckdb_httpfs/pull/7, that should make CI work properly on main, I will have CI pass / merge that one, then check CI here.

Mytherin:
Looks like CI has gone through - @carlopi good to merge?

PR #10, opened 2025-01-08
Throw HTTP error for status != 200 in InitializeMultipartUpload

Flogex:
Makes the error handling in `InitializeMultipartUpload` more consistent with `UploadBuffer` and checks for the HTTP status code before looking for the `UploadId` element. When the `UploadID` element is not found, a `HttpError` is thrown now instead of an error that is displayed as "Invalid Error".


Mytherin:
Thanks!

PR #12, opened 2025-01-21
[no-op] Resolve switch case warning

dentiny:
I included httpfs files inside of my project, but it fails compilation due to my compilation option.

Error message:
```
../../third_party/duckdb/extension/httpfs/crypto.cpp:139:1: error: control reaches end of non-void function [-Werror=return-type]
  139 | }
      | ^
cc1plus: all warnings being treated as errors
```


dentiny:
Hi @samansmink do you mind taking a look when you have some time? This PR should be an easy no-op change. Thank you!

Mytherin:
Thanks!

Issue #16, opened 2025-02-23
Data race issue when `FILE_FLAGS_PARALLEL_ACCESS`


dentiny:
(correct me if wrong), from my understanding, the PARALLEL_ACCESS flag is used the file handle created could be accessed (i.e. read) in parallel.

But check the code here
https://github.com/duckdb/duckdb-httpfs/blob/85ac4667bcb0d868199e156f8dd918b0278db7b9/extension/httpfs/httpfs.cpp#L514-L522
the only special handling for parallel access we do is disabling caching, but it's far from enough.
For example, updating the `file_offset` field is not thread-safe from multiple threads.

I think we should provide better documentation on `FILE_FLAGS_PARALLEL_ACCESS` option.

PR #17, opened 2025-02-24
Avoid update file offset in PRead interface

dentiny:
Related issues:
- https://github.com/duckdb/duckdb-httpfs/issues/19
- https://github.com/duckdb/duckdb-httpfs/issues/16

As I mentioned in the above issues, I think there're two problems with the current interface and implementation:
- Documentation isn't clear whether file offset should advance for `Read`
- Even with `PARALLEL_READ` flag provided when open file handle, current implementation suffers concurrent update with mutex protection.

This PR tries to address both issue:
- Only update offsets at `int64_t Read(&handle, buffer, nr_bytes)`
- Acquire lock when update data field under parallel access (even if it's a benign data race)

I've been working on duckdb-vfs-like filesystem for years, happy to help if duckdb team has further questions / feature requests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,8 @@ project(HTTPFsExtension)
 
 add_extension_definitions()
 
-include_directories(extension/httpfs/include ${DUCKDB_MODULE_BASE_DIR}/third_party/httplib)
+include_directories(extension/httpfs/include
+                    ${DUCKDB_MODULE_BASE_DIR}/third_party/httplib)
 
 build_static_extension(
   httpfs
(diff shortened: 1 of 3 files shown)


dentiny:
Hi @samansmink , wondering if you could take a look when you have some time? I think this is bug / needs better documentation. Thank you!

extension/httpfs/httpfs.cpp
```diff
@@ -515,22 +514,23 @@ void HTTPFileSystem::Read(FileHandle &handle, void *buffer, int64_t nr_bytes, id
 	bool skip_buffer = hfh.flags.DirectIO() || hfh.flags.RequireParallelAccess();
 	if (skip_buffer && to_read > 0) {
 		GetRangeRequest(hfh, hfh.path, {}, location, (char *)buffer, to_read);
+
+		std::lock_guard<std::mutex> lck(hfh.mu);
```
samansmink: Can we only grab the lock when `RequireParallelAccess()`? i don't think we should grab the lock when only `DirectIO()` is true, right?

samansmink (commented):
thanks for the PR! I think this looks good have 1 small comment

CMakeLists.txt
```diff
@@ -6,7 +6,8 @@ project(HTTPFsExtension)
 
 add_extension_definitions()
 
-include_directories(extension/httpfs/include ${DUCKDB_MODULE_BASE_DIR}/third_party/httplib)
+include_directories(extension/httpfs/include
```
dentiny, line 9: The CMakeList changes are made by linter, config change introduced at https://github.com/duckdb/duckdb-httpfs/pull/25

extension/httpfs/httpfs.cpp
dentiny: Both yes to your questions.

I wrote the code that way because:
- We could reduce if-else branch and code duplication;
- C++ STL mutex first spins for a while at userspace, so for short critical section like here, I don't imagine to have performance issues.

But I updated the code to only lock when necessary :)

dentiny:
> thanks for the PR! I think this looks good have 1 small comment

Thank you for the review!

dentiny:
Hi @samansmink do you mind taking another look when you have some time? I think I have addressed your comment :)

samansmink (approved):

samansmink:
looks good, thanks @dentiny!

dentiny:
Thanks for the review! I have closed the related issues.

PR #21, opened 2025-02-25
Secret auto-refresh

samansmink:
This PR implements a first version of secret refreshing.

### Problem
Currently, S3 secrets cannot be automatically refreshed. What this means is that a secret like:
```
CREATE SECRET s1 (TYPE S3, PROVIDER credential_chain);
```
Will actually call into the AWS SDK on secret creation, grab the key_id, secret, etc. and store it physically in the secret. 

This is actually kindof problematic because in various situations these credentials will actually only be temporary credentials. This is super annoying because it requires manually refreshing the secret by the user. This causes issues like https://github.com/duckdb/duckdb-aws/issues/61

### Feature
This PR adds an initial implementation of an automatic refresh mechanism. This first implementation is compatible with duckdb v1.2.0, but future implementations should probably consider improving on this by adding support for this directly in duckdb's secret manager.

The refresh mechanism works by setting the `refresh` key to true for the secret, which store all keys of the secret create statement in a separate `refresh_info` field in the secret. To refresh a secret, this field is then used to make a new create secret call.

To illustrate, lets consider the folllowing secret:

```SQL
CREATE SECRET s1 (
    TYPE S3,
    PROVIDER credential_chain,
    PROFILE 'my_profile'
    CHAIN 'config',
    REFRESH 'auto'
)
```
This statement will result in DuckDB making a KeyValue secret with the following keys:
```
- KEY_ID:  <fetched automatically by AWS SDK>
- SECRET: <fetched automatically by AWS SDK>
- SESSION_TOKEN: <fetched automatically by AWS SDK>
- REGION: <fetched automatically by AWS SDK>
- REFRESH_INFO:
     - PROFILE: my_profile
     - CHAIN: config
     - REFRESH: auto
```

The refresh mechanism is triggered whenever `S3FileHandle::Initialize` hits an HTTPException. When this happens, the secret for the corresponding path is fetched, and inspected for this `REFRESH_INFO` field. If the field exists, meaning that the secret is setup for refreshing, we will attempt to automatically re-create the secret by using the fields from the `REFRESH_INFO` field as input to what is basically a `CREATE OR REPLACE SECRET s1 ...` statement. When this succeeds, we retry the `S3FileHandle::Initialize` call.

### Testing
The refresh mechanism only really makes sense now for the `credential_chain` provider, however that live in the aws extension. To properly test this feature in this extension, i've added the option to manually pass the `REFRESH_INFO` info field. This allows testing this with the `config` provider in a meaningful way.

### TODO
This needs to be added to the aws extension for the `credential_chain` provider. This will actually make the feature useful.

This is implemented using the regular KeyValue mechanism. This allows it to work on v1.2.0. However I think it makes sense to add native support for this to the secret manager APIs which can allows us to cleanly do other stuff like refreshing secrets automatically based on expiration timestamps.

Finally: now we only support 'auto' refreshing, which is intended to mean: "let duckdb figure out the most sane thing" and is supposed to become the default value at some point. We probably want to support things like `disabled`, `on_use`, `on_error` or `on_timeout`. We can improve on this incrementally though

### DIFF
--- a/.github/workflows/MinioTests.yml
+++ b/.github/workflows/MinioTests.yml
@@ -1,0 +1,71 @@
+name: Minio Tests
+on: [push, pull_request,repository_dispatch]
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.head_ref || '' }}-${{ github.base_ref || '' }}-${{ github.ref != 'refs/heads/main' || github.sha }}
+  cancel-in-progress: true
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  minio-tests:
+    name: Minio Tests
+    runs-on: ubuntu-20.04
+    env:
+      S3_TEST_SERVER_AVAILABLE: 1
+      AWS_DEFAULT_REGION: eu-west-1
+      AWS_ACCESS_KEY_ID: minio_duckdb_user
+      AWS_SECRET_ACCESS_KEY: minio_duckdb_user_password
+      DUCKDB_S3_ENDPOINT: duckdb-minio.com:9000
+      DUCKDB_S3_USE_SSL: false
+      GEN: ninja
+      VCPKG_TARGET_TRIPLET: x64-linux
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: 'true'
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.7'
+
+      - name: Install Ninja
+        shell: bash
+        run: sudo apt-get update -y -qq && sudo apt-get install -y -qq ninja-build
+
+      - name: Setup Ccache
+        uses: hendrikmuhs/ccache-action@main
+        with:
+          key: ${{ github.job }}
+          save: ${{ github.ref == 'refs/heads/main' || github.repository != 'duckdb/duckdb' }}
+
+      - name: Setup vcpkg
+        uses: lukka/run-vcpkg@v11.1
+        with:
+          vcpkgGitCommitId: 5e5d0e1cd7785623065e77eff011afdeec1a3574
+
+      - name: Build
+        shell: bash
+        run: make
+
+      - name: Start S3/HTTP test server
+        shell: bash
+        run: |
+          cd duckdb
+          mkdir data/attach_test
+          touch data/attach_test/attach.db
+          sudo ./scripts/install_s3_test_server.sh
+          source ./scripts/run_s3_test_server.sh
+          sleep 30
+
+      - name: Write AWS credentials file
+        shell: bash
+        run: |
+          ./scripts/create_minio_credential_file.sh
+
+      - name: Test
+        shell: bash
+        run: |
+          make test
--- a/.github/workflows/MinioTests.yml
+++ b/.github/workflows/MinioTests.yml
@@ -29,7 +29,7 @@ jobs:
 
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.7'
+          python-version: '3.10'
 
       - name: Install Ninja
         shell: bash
--- a/.github/workflows/MinioTests.yml
+++ b/.github/workflows/MinioTests.yml
@@ -60,11 +60,6 @@ jobs:
           source ./scripts/run_s3_test_server.sh
           sleep 30
 
-      - name: Write AWS credentials file
-        shell: bash
-        run: |
-          ./scripts/create_minio_credential_file.sh
-
       - name: Test
         shell: bash
         run: |
--- a/.github/workflows/MinioTests.yml
+++ b/.github/workflows/MinioTests.yml
@@ -10,7 +10,7 @@ defaults:
 jobs:
   minio-tests:
     name: Minio Tests
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     env:
       S3_TEST_SERVER_AVAILABLE: 1
       AWS_DEFAULT_REGION: eu-west-1
--- a/extension/httpfs/create_secret_functions.cpp
+++ b/extension/httpfs/create_secret_functions.cpp
@@ -14,7 +14,6 @@ void CreateS3SecretFunctions::Register(DatabaseInstance &instance) {
 static Value MapToStruct(const Value &map){
 	auto children = MapValue::GetChildren(map);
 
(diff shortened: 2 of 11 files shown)


extension/httpfs/httpfs.cpp
```diff
@@ -331,7 +331,7 @@ unique_ptr<ResponseWrapper> HTTPFileSystem::GetRequest(FileHandle &handle, strin
 					    error += " This could mean the file was changed. Try disabling the duckdb http metadata cache "
 					             "if enabled, and confirm the server supports range requests.";
 				    }
-				    throw IOException(error);
```
carlopi: I can't remember if this being an IOException it's on purpose or not.

extension/httpfs/create_secret_functions.cpp
```diff
@@ -69,14 +85,74 @@ unique_ptr<BaseSecret> CreateS3SecretFunctions::CreateSecretFunctionInternal(Cli
...
+			child_list_t<Value> struct_fields;
+			for (const auto &named_param : input.options) {
+				auto lower_name = StringUtil::Lower(named_param.first);
+				struct_fields.push_back({lower_name, named_param.second});
+			}
+			secret->secret_map["refresh_info"] = Value::STRUCT(struct_fields);
+		} else if (lower_name == "refresh_info") {
+			if (refresh) {
+				throw InvalidInputException("Can not set `refresh` and `refresh_info` at the same time");
+			}
```
carlopi, line 103: should this be there also in the previous block, to check that `refresh` is not set after `refresh_info`?

extension/httpfs/create_secret_functions.cpp
```diff
@@ -69,14 +85,74 @@ unique_ptr<BaseSecret> CreateS3SecretFunctions::CreateSecretFunctionInternal(Cli
...
+
+	// TODO: change SecretManager API to avoid requiring catching this exception
+	try {
+		auto res = secret_manager.CreateSecret(context, refresh_input);
+		auto &new_secret = dynamic_cast<const KeyValueSecret&>(*res->secret);
+		DUCKDB_LOG_INFO(context, "httpfs.SecretRefresh", "Successfully refreshed secret: %s, new key_id: %s", secret_to_refresh.secret->GetName(), new_secret.TryGetValue("key_id").ToString());
+		return true;
+	} catch (InvalidInputException &e) {
+		DUCKDB_LOG_WARN(context, "httpfs.SecretRefresh", "Failed to refresh secret %s: %s", secret_to_refresh.secret->GetName(), e.what());
+		return false;
```
carlopi: I think there might be more exceptions, like stuff might fail in CreateSecret with various other problems.

extension/httpfs/s3fs.cpp
```diff
@@ -792,8 +792,32 @@ void S3FileSystem::Verify() {
...
+		HTTPFileHandle::Initialize(opener);
+	} catch (HTTPException &e) {
+		bool refreshed_secret = false;
+		auto context = opener->TryGetClientContext();
+		if (context) {
+			auto transaction = CatalogTransaction::GetSystemCatalogTransaction(*context);
+			for (const string type : {"s3", "r2", "gcs"}) {
+				auto res = context->db->GetSecretManager().LookupSecret(transaction, path, type);
+				if (res.HasMatch()) {
+					refreshed_secret |= CreateS3SecretFunctions::TryRefreshS3Secret(*context, *res.secret_entry);
```
carlopi:
```suggestion
					refreshed_secret |= CreateS3SecretFunctions::TryRefreshS3Secret(*context, *res.secret_entry);
					break;
```

carlopi (commented):
Few comments/open questions that bumped to my mind.

I am not sure how this works if like Initialization trigger extension install (via S3?) then I am not sure what happens.
And also unsure what happens in the multithreaded case? Say accessing N parquet files, I would then assume secret are refreshed N times?

extension/httpfs/httpfs.cpp
samansmink: I will restore it to make this PR safer

extension/httpfs/create_secret_functions.cpp
samansmink: correct

extension/httpfs/s3fs.cpp
samansmink: This is intentional: we will try to refresh all of the 3 secret types. It is indeed a little wonky but it is more robust like this I feel

extension/httpfs/create_secret_functions.cpp
samansmink: I changed this to ensure both IO and HTTP exceptions trigger a refresh

.github/workflows/MinioTests.yml
```diff
@@ -0,0 +1,66 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.head_ref || '' }}-${{ github.base_ref || '' }}-${{ github.ref != 'refs/heads/main' || github.sha }}
+  cancel-in-progress: true
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  minio-tests:
+    name: Minio Tests
+    runs-on: ubuntu-20.04
```
carlopi:
```suggestion
    runs-on: ubuntu-24.04
```
(ubuntu-20 will go away, either 22 or 24)

test/sql/secret/secret_refresh.test
```diff
@@ -0,0 +1,129 @@
...
+    }
+)
+
+# For now, we throw the actual error that get's thrown during refresh. Since refresh is op-in for now that ensures user can understand what's happening
+statement error
+FROM "s3://test-bucket/test-file.parquet"
+----
+Exception thrown while trying to refresh secret s1
+
+# Cleanup: drop secret and logs
```
carlopi:
```suggestion
# Cleanup: drop secret
```

carlopi (commented):
More minor nits.

I think main risk is secret lookup code going out of sync.

One think I am not sure, would cause any visible slowdown or changes in the path where:
no secret with refresh enabled exists, but still lookups are done so somehow code might act slightly differently.

Last thing, what is any lock needed in case of multi-thread Initialization of S3FileHandles?

carlopi:
Green!
Thanks for the PR / adding testing.

Moving this to be available via:
`FORCE INSTALL httpfs FROM core_nightly;`?
to get stuff rolling?

PR #25, opened 2025-03-06
Implement `RemoveFile` & `RemoveDirectory` for s3fs

### LINKED ISSUE #20: s3fs does not implement `Remove` & `RemoveDirectory` interfaces
Is it by design, or has it simply not been implemented yet?

Usecase: in pg_mooncake (https://github.com/Mooncake-Labs/pg_mooncake), we utilize duckdb and its file system to manage the storage. So `Remove` function is necessary for recycling outdated data files in s3.

YuweiXiao:
Support S3 files deletion via `DELETE` request. This feature allow systems leveraging Duckdb, such as [pg_mooncake](https://github.com/Mooncake-Labs/pg_mooncake/), to fully manage their storage files, including deletion.

fix #20

### DIFF
--- a/extension/httpfs/include/s3fs.hpp
+++ b/extension/httpfs/include/s3fs.hpp
@@ -186,6 +186,7 @@ public:
 	duckdb::unique_ptr<ResponseWrapper> PutRequest(FileHandle &handle, string s3_url, HeaderMap header_map,
 	                                               char *buffer_in, idx_t buffer_in_len,
 	                                               string http_params = "") override;
+	duckdb::unique_ptr<ResponseWrapper> DeleteRequest(FileHandle &handle, string s3_url, HeaderMap header_map) override;
 
 	static void Verify();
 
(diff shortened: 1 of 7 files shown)


YuweiXiao:
@samansmink hey, could you help review the MR? I've fixed the compilation error in the previous workflow.

samansmink (approved):
Hey @YuweiXiao, thanks for the PR! I think this looks good. 

Ideally we would add tests for this as well, but I think we are missing the infrastructure for this. Perhaps we should consider adding a `PRAGMA delete_file('...')` to duckdb so that we can actually invoke this easily in tests.

extension/httpfs/include/s3fs.hpp
```diff
@@ -19,6 +19,8 @@
 #include <exception>
 #include <iostream>
 
+#undef RemoveDirectory
```
dentiny, line 22: curious why do we need the `undef` here?

YuweiXiao: MSVC defines a same macro and would lead to clash.

dentiny: gotcha would appreciate if we could leave a comment for easier understanding :)

PR #36, opened 2025-03-28
Use HTTPException for HTTP errors in GetRangeRequest

Flogex:
To make the exceptions more consistent with the other ones thrown in `HTTPFileSystem::GetRangeRequest`, this replaces two `IOExceptions` with `HTTPExceptions`. I think using an `HTTPException` in these cases makes more sense because the error occurs during an HTTP request.


Mytherin:
Thanks! LGTM - could you just resolve the conflicts?

Flogex:
I've rebased the branch on the latest main.

Mytherin:
Thanks!

PR #41, opened 2025-04-09
Add patch for external file cache

lnkuiper:


lnkuiper:
Aha, seems like this does not check out DuckDB `main`, but `v1.2.1`, so this won't build yet.

carlopi:
Thanks a lot! Looks good to me.

PR #42, opened 2025-04-11
Bump to v1.2.2

lnkuiper:


.github/workflows/MainDistributionPipeline.yml
```diff
@@ -14,19 +14,19 @@ concurrency:
 jobs:
   duckdb-stable-build:
     name: Build extension binaries
-    uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@v1.2.1
+    uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@v1.2.2
```
carlopi, line 17: There is no 1.2.2 branch there (yet?)

I would keep this to `main`

carlopi: To be more clear, ci_tools_versions is always `main` (in all 4 places)

lnkuiper: I think we should add the v1.2.2 branch there before merging this

PR #43, opened 2025-04-13
HTTP retry on 503 status code

dentiny:
I saw discussion thread: https://github.com/duckdb/duckdb/issues/6153, which mentions 503 is a status code which indicates slow response and should be retried.

Copy from AWS official doc:
```
If you are not using an AWS SDK, you should implement retry logic when receiving the HTTP 503 error.
```
Ref: https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-design-patterns.html#optimizing-performance-timeouts-retries

Other reference docs: https://docs.digitalocean.com/products/spaces/details/limits/#:~:text=Applications%20should%20retry%20with%20exponential%20backoff%20on%20503%20Slow%20Down%20errors.
(more in the discussion thread)

### DIFF
--- a/extension/httpfs/httpfs.cpp
+++ b/extension/httpfs/httpfs.cpp
@@ -157,7 +157,6 @@ HTTPFileSystem::RunRequestWithRetry(const std::function<duckdb_httplib_openssl::
 			case 418: // Server is pretending to be a teapot
 			case 429: // Rate limiter hit
 			case 500: // Server has error
-			case 503: // Server has error
 			case 504: // Server has error
 				break;
 			default:


dentiny:
The error message doesn't look related to my change.

Mytherin:
Thanks for the PR!

I think this changeset is not correct - we are already retrying on 503.

Note that the list you are editing indicates errors *on which a retry is performed*, i.e. we retry on 408, 418, 429, 500, 503 and 504. By removing the entry from that list, we no longer retry on 503, i.e. doing the opposite of what the PR is intended to do.

dentiny:
> Thanks for the PR!
> 
> I think this changeset is not correct - we are already retrying on 503.
> 
> Note that the list you are editing indicates errors _on which a retry is performed_, i.e. we retry on 408, 418, 429, 500, 503 and 504. By removing the entry from that list, we no longer retry on 503, i.e. doing the opposite of what the PR is intended to do.

I must mis-read the `break` statement, sorry for the interruption!

PR #52, opened 2025-05-15
HTTPUtil: port changes from duckdb's #17486

carlopi:
This should be trivial, just applying patch and bumping duckdb


Mytherin:
Thanks!

PR #56, opened 2025-05-19
Absorb patch from https://github.com/duckdb/duckdb/pull/17527

carlopi:
Also bump duckdb to use latest `v1.3-ossivalis`


Mytherin:
Thanks!

PR #58, opened 2025-05-20
Replace HTTPlib client with Curl

Tmonster:
First steps to use Curl as the HTTP client. There are still some questions around when to call `curl_global_cleanup()`, since it should only be called once curl is completely done. 

I also might not be initializing options correctly, so looking for feedback on that.

### DIFF
--- a/extension/httpfs/httpfs_client.cpp
+++ b/extension/httpfs/httpfs_client.cpp
@@ -268,10 +268,10 @@ public:
 
 		auto curl_headers = TransformHeadersCurl(info.headers);
 		// transform parameters
-		auto url = info.url;
+		auto url = make_uniq<string>(info.url);
 		if (!info.params.extra_headers.empty()) {
 			auto curl_params = TransformParamsCurl(info.params);
-			url += "?" + curl_params;
+			*url += "?" + curl_params;
 		}
 
 		CURLcode res;
@@ -280,7 +280,7 @@ public:
 
 		{
 			// Set URL
-			curl_easy_setopt(*curl, CURLOPT_URL, info.url.c_str());
+			curl_easy_setopt(*curl, CURLOPT_URL, url->c_str());
 
 			// Perform HEAD request instead of GET
 			curl_easy_setopt(*curl, CURLOPT_NOBODY, 1L);
@@ -303,11 +303,14 @@ public:
 			// Execute HEAD request
 			res = curl->Execute();
 		}
-
+		Printer::Print("executed HEad");
+		Printer::Print("url is " + *url);
+		Printer::Print("body is " + result);
 
 		uint16_t response_code = 0;
 		curl_easy_getinfo(*curl, CURLINFO_RESPONSE_CODE, &response_code);
-		return TransformResponseCurl(response_code, response_header_collection ? std::move(response_header_collection) : nullptr, result, res, url);
+		Printer::Print("start transforming");
+		return TransformResponseCurl(response_code, response_header_collection ? std::move(response_header_collection) : nullptr, result, res, *url);
 	}
 
 	unique_ptr<HTTPResponse> Delete(DeleteRequestInfo &info) override {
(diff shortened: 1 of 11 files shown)


extension/httpfs/httpfs_client.cpp
```diff
@@ -2,146 +2,499 @@
...
+		if (!info.params.extra_headers.empty()) {
+			auto curl_params = TransformParamsCurl(info.params);
+			request_info->url += "?" + curl_params;
+		}
+
+		CURLcode res;
+		auto header_collector = make_uniq<HeaderCollector>();
+		{
+			// Set URL
+			curl_easy_setopt(*curl, CURLOPT_URL, request_info->url.c_str());
```
samansmink, line 271: There's now quite a bit of repetition in these curl configurations between the different request methods. Could we perhaps move the shared config between them in a function?

test/sql/copy/csv/test_csv_httpfs.test
```diff
@@ -353,3 +294,94 @@ select * from read_csv_auto('https://csvbase.com/meripaterson/stock-exchanges');
 249	North America	United States of America	Members' Exchange	NULL	2020-09-24
 250	Africa	Zimbabwe	Victoria Falls Stock Exchange	NULL	2020-11-01
 251	Asia	China	Beijing Stock Exchange	NULL	2021-12-27
+
+
+#FIXME this test fails: file is nonexistent
+mode skip
+
+query IIIIII rowsort
+SELECT * from read_csv_auto('https://github.com/duckdb/duckdb/raw/9cf66f950dde0173e1a863a7659b3ecf11bf3978/data/csv/customer.csv');
```
samansmink, line 306: Maybe its nice to use logging to ensure we get the expected http headers?

for example:
```
pragma enable_logging('HTTP');
-- do query
select response.headers['Connection'] from duckdb_logs_parsed('HTTP');
```

We could verify for example that the `__RESPONSE_STATUS__` header is properly parsed into the response headers: `select response.headers['__RESPONSE_STATUS__'] from duckdb_logs_parsed('HTTP');`

extension/httpfs/httpfs_client.cpp
```diff
@@ -2,146 +2,499 @@
...
+	// Trim trailing \r\n
+	if (!header.empty() && header.back() == '\n') {
+		header.pop_back();
+		if (!header.empty() && header.back() == '\r') {
+			header.pop_back();
+		}
+	}
+
+	// If header starts with HTTP/... curl has followed a redirect and we have a new Header,
+	// so we clear all of the current header_collection
```
samansmink: not sure I understand this, the comment suggests this is going to clear the header_collection, but then doesn't?

samansmink (commented):
Looks great! left some initial comments

carlopi:
Just a concern of mine, to keep in mind: could we have a branch (that I assumed was `main`, but could be another one) that stays in sync with what we want to release on 1.3.0?

That would mean either creating a separate `v1.3-ossivalis` branch, or the other way around have a `curl` branch.

Tmonster:
> that stays in sync with what we want to release on 1.3.0?

There is already a v1.3-ossivalis branch https://github.com/duckdb/duckdb-httpfs/tree/v1.3-ossivalis
That's what you mean? I think the plan is to merge this when green/approved to main, then ask customers who need CURL to download from Nightly.

carlopi:
Maybe we should sync in a bit, v1.3-ossivalis branch is not in sync with https://github.com/duckdb/duckdb/blob/v1.3-ossivalis/.github/config/out_of_tree_extensions.cmake, and I assume it's not compatible with v1.3-ossivalis branch of duckdb/duckdb.

Happy to have `v1.3-ossivalis` here track what gets released, but then it needs to be bumped.

Tmonster:
@carlopi I have a PR here to update the v1.3 branch to current main. We can then use that for bug fixes after this merges
https://github.com/duckdb/duckdb-httpfs/pull/63

carlopi:
I think this is now solved via https://github.com/duckdb/duckdb-httpfs/pull/96, unsure if there is something still relevant here.

carlopi:
Same, I think solved via https://github.com/duckdb/duckdb-httpfs/pull/96, thanks

PR #59, opened 2025-05-20
Update test_secret_type.test: explicit load

carlopi:
Otherwise test fails in autoloading mode


carlopi:
I think this should be safe, otherwise failures are like: https://github.com/carlopi/duckdb/actions/runs/15137997505/job/42558217288#step:6:23850

PR #64, opened 2025-05-26
[WIP] HTTPMetadataCache should also store OpenFileExtendedInfo

Tmonster:
If there is extended info, options like "validate_external_cache" can be read from any thread. 

This is WIP because the extended_info is still a shared pointer between the HTTPMetadataCache and HTTPFileHandle etc. It's still possible to have two handles reading from the same memory. The problem with that is what happens if the extended_options object becomes very big? (We will probably deal with that later then).

Eventually there may be extra options, one I can see is a "direct_get" option, which will translate to not executing a head request and executing a GET request. This will help with Lakehouse formats (i.e Iceberg) where we don't want to execute a HEAD request if we KNOW we are still going to execute a GET request.


lnkuiper:
Superseded by https://github.com/duckdb/duckdb-httpfs/pull/427 (I'll send another PR to clean it up a bit once https://github.com/duckdb/duckdb/pull/25426 is merged)

PR #67, opened 2025-06-05
Allow complete downloads to happen without first making a GET request

Tmonster:
In case users know they want to download the file completely without making HEAD requests to learn the size of the file. This will be handy for Iceberg and avro reads.

Don't know about the name direct_complete download. Happy to change if needed.


Tmonster:
Is main still the branch for v1.3.1? Or has main diverged?

samansmink (approved):
LGTM, maybe the option could be `force_full_download`?  no strong opinion though

edit: given that we already have a FullDownload functionality

Tmonster:
I can change it, force_full_download sounds better

PR #71, opened 2025-06-23
Fixup syntax error + temporary skip wasm targets

carlopi:
Workflow changes is since wasm work is currently carried forward in the `wasm` branch, and it's not clear yet how to reconcile the different set of dependencies.

Formatting error fix is that otherwise on formatting the exception message that would turn into an more heavy InternalException.


Tmonster (approved):
lgtm 👍

PR #74, opened 2025-07-03
Add code quality workflows: format and tidy

carlopi:


carlopi:
While this PR is ready, it's probably worth waiting a bit longer to first absorb the rest of the work.

Makefile
```diff
@@ -8,3 +8,23 @@ CORE_EXTENSIONS=''
 
 # Include the Makefile from extension-ci-tools
 include extension-ci-tools/makefiles/duckdb_extension.Makefile
+
+#### Misc
+format-check:
+	python3 duckdb/scripts/format.py --all --check --directories extension/httpfs test
```
Mytherin, line 14: Maybe we can just move the extension code to `src` like other extensions?

carlopi: Yes, moving to `src` is right, I will do so post 1.3.2 given there are a bunch of in-flight PRs.

Also format code might need to take the paths from a Makefile variable instead of hardcoding, but that's minor.

Tmonster:
I think this happens via the extension-ci-tools repo by default right? So this can also be closed?

carlopi:
No, this needs to be added on a repo-by-repo basis, at the very least adding something like:
```

  code-quality:
    name: Code Quality
    uses: duckdb/extension-ci-tools/.github/workflows/_extension_code_quality.yml@main
    with:
      extension_name: httpfs
      duckdb_version: v1.3-ossivalis
      ci_tools_version: main
      extra_toolchains: 'python3'
      format_checks: 'format;tidy'
```
or equivalent (maybe changing branch names)

PR #76, opened 2025-07-03
Wasm HTTPUtil on main

carlopi:
This is not super pretty, since I have not found a more proper way to feature detect HTTPFSClient than `#ifdef`'s on `EMSCRIPTEN` being defined.
Weakest part of this PR are the changes in [extension/httpfs/httpfs_client.cpp](https://github.com/duckdb/duckdb-httpfs/compare/main...carlopi:duckdb_httpfs:wasm_httputil_on_main?expand=1#diff-672a3be70016b1bbb6512e78cc15494649f9d016d48041eee803576d91f3498b), open to suggestions.

EDIT: I did changed approach, and skip the file at the CMake level

After this PR `httpfs` is re-enabled for all platforms.

### DIFF
--- a/extension_config.cmake
+++ b/extension_config.cmake
@@ -15,4 +15,5 @@ duckdb_extension_load(httpfs
 	SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}
 	INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/extension/httpfs/include
 	${LOAD_HTTPFS_TESTS}
+	LINKED_LIBS "../../third_party/mbedtls/libduckdb_mbedtls.a"
 )
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -19,7 +19,6 @@ jobs:
       extension_name: httpfs
       duckdb_version: v1.3-ossivalis
       ci_tools_version: main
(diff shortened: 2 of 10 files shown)


carlopi:
This is ready to be merged on my side

extension_config.cmake
```diff
@@ -15,4 +15,5 @@ duckdb_extension_load(httpfs
 	SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}
 	INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/extension/httpfs/include
 	${LOAD_HTTPFS_TESTS}
+	LINKED_LIBS "../../third_party/mbedtls/libduckdb_mbedtls.a"
```
Mytherin, line 18: Is this necessary here?

carlopi: Unfortunately, yes, I have not found a clean way to clean this up.

It should picked up from CMake, but that requires some changes in duckdb/duckdb that are not there (and unsure if they can land today).

carlopi: I can fix this up, question is: it's a blocker?

Mytherin: I don't know what problems it causes but it seems hacky - maybe it's fine. I'll leave it up to you

carlopi: It does smell badly, you are right, now triggered to find the actual fix.

PR #78, opened 2025-07-03
Continue in S3FileHandle::Initialize on secret refresh

Mytherin:
The intention here is to re-run the same method using the refreshed credentials. Currently we are calling only the base-class method again and then returning. This skips a bunch of code related to uploading - which could potentially cause issues. This PR inverts the condition to instead throw if there is no secret refresh - and otherwise refresh and continue.


samansmink (approved):
LGTM, good catch!

PR #77, opened 2025-07-03
Use curl as optional client

carlopi:
This integrates https://github.com/duckdb/duckdb-httpfs/pull/58 AND https://github.com/duckdb/duckdb-httpfs/pull/76, since they have some interactions that needed some care.

After https://github.com/duckdb/duckdb/pull/18107 landed in duckdb/duckdb, and moving the `duckdb` submodule to a recent commit on `v1.3-ossivalis`, this PR allows to switch at runtime based on the newly added `httpfs` config option `httpfs_client_implementation`:

```sql
D SET logging_storage=stdout;
D PRAGMA enable_logging('HTTP');
D SET httpfs_client_implementation='default';
D select count(*)
  FROM 'https://github.com/duckdb/duckdb-data/releases/download/v1.0/event_baserunning_advance_attempt.parquet';
[LOG] 2025-07-03 10:06:18.479, HTTP, DEBUG, {'request': {'type': HEAD, 'url': 'https://github.com/duckdb/duckdb-data/releases/download/v1.0/event_baserunning_advance_attempt.parquet', 'headers': {}}, 'response': {'status': OK_200, 'reason': OK, 'headers': {X-Timer='S1751537178.169255,VS0,VE1', x-ms-request-id=91401b99-a01e-000c-7caa-dd765d000000, X-Served-By='cache-iad-kcgs7200132-IAD, cache-rtm-ehrd2290029-RTM', Fastly-Restarts=1, x-ms-lease-status=unlocked, x-ms-creation-time='Tue, 17 Jan 2023 21:28:40 GMT', x-ms-blob-type=BlockBlob, x-ms-blob-content-md5='0OgVgULsCWfa1mU4BlFEbg==', X-Cache-Hits='3730, 1', Via='1.1 varnish, 1.1 varnish', Server=Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0, x-ms-version=2025-05-05, X-Cache='HIT, HIT', ETag='"0x8DAF8D1CD43CA79"', Connection=keep-alive, x-ms-lease-state=available, Last-Modified='Tue, 17 Jan 2023 21:28:40 GMT', Date='Thu, 03 Jul 2025 10:06:18 GMT', Content-Length=21916382, Content-Type=application/octet-stream, Content-Disposition='attachment; filename=event_baserunning_advance_attempt.parquet', x-ms-server-encrypted=true, Age=806, Accept-Ranges=bytes}}}, CONNECTION, 2, 11, NULL
┌────────────────┐
[...]

PR #80, opened 2025-07-03
Add httpfs_client_implementation option, currently only `default` and `httplib` are valid

carlopi:
Idea of this is allowing, in future versions of `httpfs` to offer a different version, namely `curl`, and possibly move the default.

Infrastructure would allow to register already in duckdb the option.

### DIFF
--- a/extension/httpfs/httpfs_extension.cpp
+++ b/extension/httpfs/httpfs_extension.cpp
@@ -12,6 +12,25 @@
 
 namespace duckdb {
 
+static void SetHttpfsClientImplementation(DBConfig &config, const string &value) {
+		if (config.http_util && config.http_util->GetName() == "WasmHTTPUtils") {
+			if (value == "wasm" || value == "default") {
+				// Already handled, do not override
+				return;
+			}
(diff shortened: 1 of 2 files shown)


carlopi:
My idea is that this could go into 1.3.2, and then post 1.3.2 we can consider merging #58 (more ready form in #77), adding `curl` as alternative implementation, and have users try it out / give feedback already via:
```
SET httpfs_client_implementation = curl;
```

Then in a follow up / after relevant testing we could consider changing the default.

Tmonster (approved):
Thanks. LGTM

carlopi:
Question is like: merge + add to 1.3.2 or to be done later?

Tmonster:
I don't really see it adding any functionality that is necessary for v1.3.2, but it also seems like a relatively safe PR, so not against it. If I had to choose, I would say wait until after v1.3.2 then merge it so it is in CORE_NIGHTLY

carlopi:
Yes, it's minor, only minor advantage is the fact that setting will be there in any version of httpfs in 1.3.2, vs being an error in some cases and present in others, but I can see also the call for that, and that's your call.

PR #93, opened 2025-08-03
Add OAuth2 bearer token support for GCS

danklynn:
Apache Polaris vends an OAuth2 access token for Iceberg tables hosted in GCS. This PR adds support for fetching files from GCS using these bearer tokens. Tested with read_parquet and iceberg_scan.

### DIFF
--- a/extension/httpfs/create_secret_functions.cpp
+++ b/extension/httpfs/create_secret_functions.cpp
@@ -114,6 +114,10 @@ unique_ptr<BaseSecret> CreateS3SecretFunctions::CreateSecretFunctionInternal(Cli
 											lower_name, named_param.second.type().ToString());
 			}
 			secret->secret_map["requester_pays"] = Value::BOOLEAN(named_param.second.GetValue<bool>());
+		} else if (lower_name == "bearer_token" && input.type == "gcs") {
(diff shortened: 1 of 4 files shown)


Tmonster:
Hi.

Thanks for the PR!

So If I understand this correctly, Polaris is returning a OAuth2 token to be used when reading from a GCS bucket? I don't think the oauth2 credentials will be picked up when attaching to the iceberg rest catalog, but we should be able to add this soon.

Otherwise LGTM. Thanks for the tests.

danklynn:
Correct. Polaris vends credentials in the loadtable response if the `X-Iceberg-Access-Delegation: vended-credentials` header is provided in the request:

```json
{
  "metadata" : ...,
  "metadta-location" : ...,
  "config" : {
    "gcs.oauth2.token": "{OAUTH2 BEARER TOKEN}",
    "gcs.oauth2.token-expires-at": "{milliseconds since epoch}"
  }
}
```

If we can get this merged, I can try to add support to the iceberg extension to support this functionality and pass the bearer token to httpfs. It would be great if we could get this in before the 1.4 freeze

Tmonster:
@samansmink this is good to go from my side. I've been able to test it locally with an actual GCS setup

PR #94, opened 2025-08-04
[minor] Place functions into anonymous namespace

dentiny:
I use the same function name `TimestampToTimeT` under `duckdb` namespace, met some linking issue and found we already have the same function under the same namespace at httpfs.

This PR is a no-op trivial change to place functions only used in impl file into an anonymous namespace, to scope it only visible from the same translation unit.

Follows PR in ducdb main repo: https://github.com/duckdb/duckdb/pull/18177

### DIFF
--- a/extension/httpfs/httpfs.cpp
+++ b/extension/httpfs/httpfs.cpp
@@ -21,6 +21,53 @@
 
 namespace duckdb {
 
+namespace {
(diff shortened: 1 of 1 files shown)


Tmonster (approved):
Thanks! 
LGTM

lnkuiper:
This function is no longer there

PR #99, opened 2025-08-22
Bugfix for headers in S3 requester pays mode - when session token is present. 

prmoore77:
Bugfix for S3 requester pays mode.  Per [AWS Documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html) the headers must be sorted.  When a session token is present, we have a bug that causes failure when requester pays mode is also used.

Thanks to @wiamsuri for finding and reporting the issue!


samansmink:
Thanks for the fix @prmoore77!

PR #105, opened 2025-09-01
Append content-type only on DuckDB-Wasm

carlopi:
I feel this is not the right solution, BUT reverting https://github.com/duckdb/duckdb-httpfs/pull/101/commits/7dfdcf05df3e46c9d3cf595daff7824732ae797d is the right first step, given otherwise it broke signing to S3

### DIFF
--- a/extension/httpfs/s3fs.cpp
+++ b/extension/httpfs/s3fs.cpp
@@ -75,7 +75,9 @@ static HTTPHeaders create_s3_header(string url, string query, string host, strin
 	hash_str canonical_request_hash_str;
 	if (content_type.length() > 0) {
 		signed_headers += "content-type;";
+#ifdef EMSCRIPTEN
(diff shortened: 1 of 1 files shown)


samansmink:
Won't this still break things on wasm?

carlopi:
This I added to have PUT working on the Wasm side.

My understanding is that content-type is somewhat of a special header, in the sense that both our httplib AND potentially some remote servers (AWS) will interpret his absence in funny and benevolent ways.

I think the actual solution would be to pass an empty "content-header" to PUT requests, but I have not tested that properly.

carlopi:
My thinking is: that line was added for Wasm, in the case of Wasm as far as I am aware it's needed, in native does break stuff, we don't really have time to properly stress test this (I am not confident on making a call), so maybe IFDEF it is.

PR #106, opened 2025-09-01
Single PUT when uploading single buffer-files to S3

carlopi:
We currently follow always same path on uploading to S3-like storage:
* 1 POST to start upload
* 1 or more PUT requests to upload buffers
* 1 POST to finalize upload

While this is general and battle tested, there is an improvement, we could check the count of buffers to be uploaded, and if that is at 1 we could perform a single PUT (moving from 3 requests over the network to a single one).

This is particularly significant for writes to data lakes in general and Iceberg in particular, due to the fact that both JSON and AVRO files have to be uploaded (and that means a `INSERT INTO <table> VALUES ()` costs currently 10+ sequential network requests)

Next up: figuring out how to cut the HEAD request currently performed when opening a FileHandle, that could properly move to single request.

For the reviewer(s): I am assuming we always fully upload a file, and never just a single part of an existing file. It would be much better to check that explicitly, unsure if there is else needed.

### DIFF
--- a/extension/httpfs/s3fs.cpp
+++ b/extension/httpfs/s3fs.cpp
@@ -299,8 +299,8 @@ S3ConfigParams S3ConfigParams::ReadFrom(optional_ptr<FileOpener> opener) {
 void S3FileHandle::Close() {
 	auto &s3fs = (S3FileSystem &)file_system;
 	if (flags.OpenForWriting() && !upload_finalized) {
-		s3fs.FlushAllBuffers(*this);
-		if (parts_uploaded) {
+		bool did_upload = s3fs.FlushAllBuffers(*this);
+		if (parts_uploaded && did_upload) {
 			s3fs.FinalizeMultipartUpload(*this);
 		}
 	}
@@ -336,6 +336,8 @@ string S3FileSystem::InitializeMultipartUpload(S3FileHandle &file_handle) {
 
 	open_tag_pos += 10; // Skip open tag
 
+	file_handle.initialized_multipart_upload = true;
+
 	return result.substr(open_tag_pos, close_tag_pos - open_tag_pos);
 }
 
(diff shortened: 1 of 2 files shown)


extension/httpfs/s3fs.cpp
```diff
@@ -403,7 +418,7 @@ void S3FileSystem::UploadBuffer(S3FileHandle &file_handle, shared_ptr<S3WriteBuf
 	// Free up space for another thread to acquire an S3WriteBuffer
 	write_buffer.reset();
 
-	NotifyUploadsInProgress(file_handle);
+	//NotifyUploadsInProgress(file_handle);
```
samansmink, line 421: nit: uncomment?

samansmink (commented):
Great idea! as agreed while discussing this in person: we should expand the httpfs ci before merging this

carlopi:
Leaving a comment for later:
```
        bool use_sse_kms = auth_params.kms_key_id.length() > 0 && (method == "POST" || method == "PUT") &&
                           query.find("uploadId") == std::string::npos;
```
code-path needs also to be looked at in light of this

extension/httpfs/s3fs.cpp
```diff
@@ -895,7 +924,8 @@ void S3FileHandle::Initialize(optional_ptr<FileOpener> opener) {
 		            Storage::DEFAULT_BLOCK_SIZE;
 		D_ASSERT(part_size * max_part_count >= config_params.max_file_size);
 
-		multipart_upload_id = s3fs.InitializeMultipartUpload(*this);
+		initialized_multipart_upload = false;
+		//multipart_upload_id = s3fs.InitializeMultipartUpload(*this);
```
samansmink, line 928: nit: remove altogether?

samansmink (approved):
LGTM!

samansmink:
Before merging though, we should probably add some tests here like https://github.com/duckdb/duckdb-httpfs/pull/113/files#diff-6eb7d9e04e92888e7c8615ffe9fab2e708520b65455c7225a7a3833b419037ce that confirm this works as expected

carlopi:
Thanks for the review, I forgot I had a better version of this lying around (see https://github.com/duckdb/duckdb-httpfs/pull/128), closing since I believe that one is better / more proper

PR #109, opened 2025-09-03
Update main to use duckdb v1.4

Tmonster:

PR #110, opened 2025-09-03
Merge main into v1.4-andium

Tmonster:
Merges main into v1.4-andium. Then we can merge v1.4 into main

Unblocks https://github.com/duckdb/duckdb-httpfs/pull/109


carlopi:
@Tmonster: I think https://github.com/duckdb/duckdb-httpfs/pull/105, or some equivalent, needs in, otherwise connection to AWS's S3 is broken due to extra content type

PR #114, opened 2025-09-05
When a HTTPFIleHandle has an Etag, and that changes over time, throw an error

carlopi:
Basically same as https://github.com/duckdb/duckdb-httpfs/pull/111, adding an option to disable the behaviour.
```
SET unsafe_disable_etag_checks = true;
```
Brings back current default, that is not checking, to avoid some yet unknown relevant workflow that will start failing due to this change.

### DIFF
--- a/extension/httpfs/include/httpfs_client.hpp
+++ b/extension/httpfs/include/httpfs_client.hpp
@@ -21,6 +21,7 @@ struct HTTPFSParams : public HTTPParams {
 	idx_t hf_max_per_page = DEFAULT_HF_MAX_PER_PAGE;
 	string ca_cert_file;
 	string bearer_token;
+	bool unsafe_disable_etag_checks {false};
 	shared_ptr<HTTPState> state;
 };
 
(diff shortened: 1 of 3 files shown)


extension/httpfs/include/httpfs_client.hpp
```diff
@@ -21,6 +21,7 @@ struct HTTPFSParams : public HTTPParams {
 	idx_t hf_max_per_page = DEFAULT_HF_MAX_PER_PAGE;
 	string ca_cert_file;
 	string bearer_token;
+	bool unsafe_disable_etag_checks {false};
```
Tmonster, line 24: Is this still needed? I think the setting is read from the db settings right?

Tmonster (approved):
just the one comment, but not a blocker.

extension/httpfs/include/httpfs_client.hpp
carlopi: Path is like:
1. read from db settings
2. saved in HTTPFSParams
3. passed around as HTTPFSParams
4. read from there

So I think this is needed.

PR #117, opened 2025-09-08
Set User-Agent to DuckDB's custom db.config.UserAgent()

carlopi:


carlopi:
This has been tested with a local http-server + enabling logging of incoming headers.

PR #123, opened 2025-09-11
Wasm fixes: move Curl to separate header (so curl is not needed to be linked in)

carlopi:
Should be a NOP for native, while skipping compexity (and, if misused, crashes) on the Wasm side

### DIFF
--- a/extension/httpfs/include/httpfs_curl_client.hpp
+++ b/extension/httpfs/include/httpfs_curl_client.hpp
@@ -1,0 +1,58 @@
+#pragma once
+
+#include "duckdb/common/http_util.hpp"
+
+namespace duckdb {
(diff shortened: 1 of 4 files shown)


extension/httpfs/include/httpfs_curl_client.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include "duckdb/common/http_util.hpp"
+
+namespace duckdb {
+class HTTPLogger;
+class FileOpener;
+struct FileOpenerInfo;
+class HTTPState;
+
+#include <curl/curl.h>
```
dentiny, line 11: Usually we don't include header file in a namespace, since all symbols inside of the header file is extra namespace scoped.

PR #125, opened 2025-09-16
bump to v1.4.0

c-herrewijn:
bump to v1.4.0


carlopi:
Thanks, I think this I will do manually, first bump 1.4-andium, then also main.

c-herrewijn:
FYI: I'll sit with Sam tomorrow to improve the logic in case extension repositories (also) have a release branch.

PR #126, opened 2025-09-16
Remove s3 region default, check error message for Unknown Error to HTTP HEAD

Tmonster:
Temporary workaround for https://github.com/duckdblabs/duckdb-internal/issues/5905

@samansmink unsure if v1.4-andium is now the default bug fix branch.

Seems like httplib.hpp doesn't expect `HEAD` requests to get redirects. At this line is the check for a redirect
https://github.com/duckdb/duckdb/blob/d19df69e14a4e7bdce94ebe5fd9f398bbbe30c57/third_party/httplib/httplib.hpp#L7405

This error manifests in a way where the request fails, and the response information is not saved in any way, so we cannot  parse any of the response information in httpfs. Hence also the message "unknown error"

This is a pr to remove the default of us-east-1, which already clears a lot of things up. The Legacy Aws s3 endpoint will be used, which will work to read data (although not in test infrastructure, unclear why). The issue will be further be cleaned up in v1.4.1 with a change to the httplib code to fix the redirects on HEAD calls

This seems to also include a bug fix for requester-pays, but I think that will fall away if main is merged into v1.4. I am unsure what the current process is for bug fixes for v1.4 in httpfs

### DIFF
--- a/extension/httpfs/httpfs_extension.cpp
+++ b/extension/httpfs/httpfs_extension.cpp
@@ -62,7 +62,7 @@ static void LoadInternal(ExtensionLoader &loader) {
 	config.AddExtensionOption("ca_cert_file", "Path to a custom certificate file for self-signed certificates.",
 	                          LogicalType::VARCHAR, Value(""));
 	// Global S3 config
-	config.AddExtensionOption("s3_region", "S3 Region", LogicalType::VARCHAR, Value("us-east-1"));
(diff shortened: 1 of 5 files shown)


Tmonster:
Closing, this is not the correct approach.
If httputil does not see a "location" header in a 301 response, then it doesn't return any response headers.
The response headers include the correct region though. I plan to return that information to httpfs. Then in httpfs we can parse the correct region and make a request to that region. Potentially adding some log functionality along the way as well

PR #133, opened 2025-09-25
Move sources to src/ directory

Mytherin:
The current source files are in a non-standard directory (`extension/httpfs`). This breaks a bunch of assumptions made by e.g. `extension-ci-tools` - causing the format script to not work correctly. This PR moves them to the `src` directory to be consistent with the rest of the extensions.


carlopi (approved):
Thanks!

PR #135, opened 2025-09-25
Merge v.14 into main

Mytherin:


.github/workflows/MainDistributionPipeline.yml
```diff
@@ -28,6 +28,7 @@ jobs:
     secrets: inherit
     with:
       extension_name: httpfs
-      duckdb_version: v1.4-andium
+      duckdb_version: v1.4.0
       ci_tools_version: main
-      deploy_latest: ${{ startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main' }}
+      deploy_latest: ${{ startsWith(github.ref, 'refs/heads/v') || github.ref == 'refs/heads/main' }}
```
carlopi: I think this should be changed to:
```suggestion
      deploy_latest: ${{ startsWith(github.ref, 'refs/heads/v') }}
```

Idea is that if both v1.4-andium and main build vs duckdb v1.4.0 (and this is good), then only one should publish core_extensions, since otherwise last merged PR will override what's installed via `INSTALL httpfs FROM core_nightly`. I think `main` is still reachable via versioned.

This is not a blocker, I can think how to clean this up properly (it's true for most extensions) but once I noted this I can't unseen this.

carlopi:
Thanks!

PR #138, opened 2025-09-27
Check libcurl return value

dentiny:
This PR does two things:
- (major) Assert libcurl operations' return value, otherwise it could fail silently
- (major) Move header file inclusion out of `duckdb` namespace
- (minor) Use atomic variable for httpfs client count, so we could (1) reduce one global variable; (2) slightly better perf


src/include/httpfs_curl_client.hpp
```diff
@@ -2,13 +2,13 @@
 
 #include "duckdb/common/http_util.hpp"
 
+#include <curl/curl.h>
```
dentiny, line 5: We shouldn't include header files inside of namespace, otherwise all symbols inside are namespace scoped.

carlopi: You are right, thanks!

dentiny: Thanks for the quick reply! This PR somehow fails unit test (which I haven't checked, will do later), so put a separate PR for header inclusion here: https://github.com/duckdb/duckdb-httpfs/pull/139

carlopi:
I think first point is valid, unsure if it should be a D_ASSERT (meaning only there on relassert) or then a throw of some kind (unsure which), so that is visible to all users.

Second split-off PR has already been merged.

Third I guess so.

dentiny:
> I think first point is valid, unsure if it should be a D_ASSERT (meaning only there on relassert) or then a throw of some kind (unsure which), so that is visible to all users.

The reason I use check failure here is: we're simply setting configs for curl handle, and they're not expected to fail.

dentiny:
Current implementation is actually buggy:
- Under release debug mode, `D_ASSERT` is translated to `assert`, which [does nothing](https://github.com/llvm/llvm-project/blob/67141c74272838919985ce1931c42365b1790c6a/libc/src/assert/assert.h#L18-L20)
- So to me, the best practice is never place any real logic inside of `D_ASSERT`

PR #139, opened 2025-09-27
Fix header inclusion

dentiny:
Header files shouldn't be included in the namespace, otherwise all symbols inside are namespace scoped.


carlopi:
Thanks for correcting this.

PR #140, opened 2025-09-27
[minor] Remove unused function

dentiny:
I found this function is unused anywhere:
```sh
[/tmp/duckdb-httpfs] (main) 
vscode@eebee01fa8d2$ git grep "SetHttpfsClientImplementation"
src/httpfs_extension.cpp:static void SetHttpfsClientImplementation(DBConfig &config, const string &value) {
```

### DIFF
--- a/src/httpfs_extension.cpp
+++ b/src/httpfs_extension.cpp
@@ -15,25 +15,6 @@
 
 namespace duckdb {
 
-static void SetHttpfsClientImplementation(DBConfig &config, const string &value) {
-	if (config.http_util && config.http_util->GetName() == "WasmHTTPUtils") {
-		if (value == "wasm" || value == "default") {
-			// Already handled, do not override
-			return;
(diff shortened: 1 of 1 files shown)


src/httpfs_extension.cpp
```diff
@@ -15,25 +15,6 @@
 
 namespace duckdb {
 
-static void SetHttpfsClientImplementation(DBConfig &config, const string &value) {
```
dentiny, line 18: Looks duplicate functionality with https://github.com/duckdb/duckdb-httpfs/blob/6b683b1d0b8df3596c65fea97952425ec9469d9a/src/httpfs_extension.cpp#L99-L125

dentiny:
Hi @carlopi do you mind taking a look? I think it should be fairly simple and straightforward, thanks!

carlopi:
Thanks!

PR #143, opened 2025-10-01
Add allow_asterisks_in_http_paths setting, that default to false, so …

carlopi:
…that Globs on HTTP file system will now throw

This fix an existing `FIXME` in the code, with what I believe is a sensible default.

This came up connected to running queries across different backends, and failure mode would be different, while the expected behaviour (I think) is no globs on `http://` or `https://`.

Old behaviour can be enabled via `SET allow_asterisks_in_http_paths = true;` (and arguably, if this is to land now, default needs to be swapped)

### DIFF
--- a/src/httpfs_extension.cpp
+++ b/src/httpfs_extension.cpp
@@ -62,6 +62,8 @@ static void LoadInternal(ExtensionLoader &loader) {
 	    "http_keep_alive",
 	    "Keep alive connections. Setting this to false can help when running into connection failures",
 	    LogicalType::BOOLEAN, Value(true));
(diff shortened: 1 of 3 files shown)


carlopi:
I like this, question on the default, unsure if going bold or keep it.

Reasons for forbid by default: it's reasonable, and you will get an error message so you know how to act

Reason to keep current default: no changes

samansmink (commented):
I like this, I don't think this should go in v1.4-andium though. This might break peoples workflow (albeit with an easy fix), so it should be pushed to v1.5.0

samansmink (approved):

PR #146, opened 2025-10-09
Fix CSV COPY to S3 using FILE_SIZE_BYTES

### LINKED ISSUE #18381: CSV - COPY TO doesn't respect the FILE_SIZE_BYTES option
### What happens?

Differently from parquet, exporting with CSV doesn't respect the option FILE_SIZE_BYTES, it still exports a single .csv document with the whole content, regardless of the limit given.

### To Reproduce

tested with the python client

### OS:

linux x86_64

### DuckDB Version:

1.3.0

### DuckDB Client:

python

### Hardware:

_No response_

### Full Name:

Dean De Leo

### Affiliation:

RelationalAI

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

Not applicable - the reproduction does not require a data set

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

BrynjarM:
FILE_SIZE_BYTES did not influence CSV COPY operations to S3 since, unlike the filesystem variant, the file length was consistently reported as 0. The implementation now updates the length as data is written + test added (based on https://github.com/duckdb/duckdb/blob/main/test/sql/copy/file_size_bytes.test).

Fixes https://github.com/duckdb/duckdb/issues/18381


BrynjarM:
@Tmonster, @carlopi, @samansmink
I'm bringing this pull request to your attention as I'm unable to formally request reviewers. I noticed you are all active contributors and was hoping you might have a moment to look.
My primary goal with this PR was to clearly illustrate the issue from the bug report (that I found after running into this exact issue). As I'm new to the codebase, please feel free to take a completely different approach to the solution.
The key outcome is to ensure the `FILE_SIZE_BYTES` setting is correctly respected when writing CSVs to S3, allowing us to control the number of output files.
Thank you for your time and consideration.

taniabogatsch:
Thanks for the PR! We'll take a look at the change in the coming days 👍

Tmonster:
Hi @BrynjarM , 

Looks good to me after some local testing. I noticed this doesn't work for Parquet files, which according to the [documentation it should](https://duckdb.org/docs/stable/sql/statements/copy#copy--to-options). Is this something you tested? 

I will double check internally what we expect here, sometimes our documentation is wrong.

BrynjarM:
> Hi @BrynjarM ,
> 
> Looks good to me after some local testing. I noticed this doesn't work for Parquet files, which according to the [documentation it should](https://duckdb.org/docs/stable/sql/statements/copy#copy--to-options). Is this something you tested?
> 
> I will double check internally what we expect here, sometimes our documentation is wrong.

The parquet logic is implemented differently and works just fine in the current version.
It's only the csv that was having issues as it's realying on the written size.

Tmonster:
My testing shows the parquet logic not working, but that is also the case in main, so we don't have to worry about it here

Tmonster (approved):

carlopi:
And thanks!

PR #147, opened 2025-10-13
[Tests] Move test cases to the main repo

ZuleykhaPavlichenkova-TomTom:
This should fix the issue when the unit tests runner uses test cases from the extensions dependencies and cannot find the file by the path from the test case because the `__TEST_DIR__` is the same as `__WORKING_DIR__` ([see this](https://github.com/duckdb/duckdb/blob/192fbda28ddae01a675a8c41a15f3bef5a68395b/test/sqlite/test_sqllogictest.cpp#L66)) 

UPD: by moving problematic test cases to the `duckdb/duckdb` repo as @carlopi suggested below
Mirroring PR in `duckdb/diuckdb` - https://github.com/duckdb/duckdb/pull/19368

~~The downside of this PR is that you'll have to provide the `DUCKDB_DATA_DIR='duckdb' ` (or they'll be skipped) for current repo and `DUCKDB_DATA_DIR=`pwd` ` for `duckdb/duckdb` repo.~~


Tmonster (commented):
Is there a default value for `DUCKDB_DATA_DIR`, Also maybe the env variable should include the `data` dir if it is called `DUCKDB_DATA_DIR`?. So in the tests in this PR, the value would be `duckdb/data`

I feel like we can pass the value `duckdb/data` for the tests here in httpfs right?

ZuleykhaPavlichenkova-TomTom:
> Is there a default value for `DUCKDB_DATA_DIR`

There is no default value, but we can use `duckdb` as a default value. 

> `duckdb/data` for the tests here in httpfs right

then in `duckdb/duckdb` we will have to assign "`pwd`/data" to this variable
And I don't see, why we can't, but also don't see a real necessity of that.

Tmonster:
Has there already been a PR merged in the duckdb repo that initialized this variable for tests that require it?

ZuleykhaPavlichenkova-TomTom:
> Has there already been a PR merged in the duckdb repo that initialized this variable for tests that require it?

No, because I thought this needs to be merged first, otherwise it still be failing in `duckdb/duckdb`

carlopi:
I don't think this is the correct solution, maybe just move the 2 test files back to duckdb/duckdb?

ZuleykhaPavlichenkova-TomTom:
@carlopi can we already merge this PR? Please?

carlopi:
Thanks!

Issue #151, opened 2025-10-27
CMake: Incorrect link order for static OpenSSL and Curl dependencies in httpfs extension


ed-w-ds:
When building DuckDB with the httpfs extension and linking against static versions of libcurl and OpenSSL (libcurl.a, libssl.a, libcrypto.a), linker errors occur.

The build fails with undefined reference errors for OpenSSL symbols (e.g., OCSP_basic_verify, SSL_alert_desc_string_long) that are required by libcurl's OpenSSL backend functions.

This happens because the target_link_libraries commands in the httpfs extension's CMakeLists.txt link the OpenSSL libraries before the libcurl library for the httpfs_extension and httpfs_loadable_extension targets.
When linking static libraries, the linker generally requires that libraries providing symbols appear after the libraries that use those symbols on the final link command line. The current CMake order results in an incorrect linker command order, causing the linker to fail to resolve symbols needed by libcurl.a.

This is easily solved by linking curl before openssl in `CMakeLists.txt` file.

carlopi:
Solved via https://github.com/duckdb/duckdb-httpfs/pull/152

PR #152, opened 2025-10-28
Fix CMake link order for static OpenSSL and Curl dependencies

ed-w-ds:
Fixes issue #151


carlopi:
Thanks, this looks to make sense to me.

PR #157, opened 2025-11-05
Add s3 access grants feature

alisoleimani:
S3 access grants(https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-grants.html) is a solution to control access to s3 objects at scale. It is used by some big companies to control access to s3 objects. 
We need this feature to be able to use our data lake.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -20,6 +20,7 @@
 #include "duckdb/common/multi_file/multi_file_list.hpp"
 
 #include "create_secret_functions.hpp"
+#include "lru_cache.hpp"
 
 #include <iostream>
 #include <thread>
@@ -29,9 +30,22 @@
 
 namespace duckdb {
 
+struct TemporaryAWSCredential
+{
+	string access_key_id;
+	string secret_access_key;
+	string session_token;
+	timestamp_t expiration;
+};
+
+static LRUCache<string,string> AccountIdCache(1024);
+static LRUCache<string,string> BucketOwnerAccountIdCache(2048);
+static LRUCache<string,TemporaryAWSCredential> AccessGrantsCache(4096);
+static LRUCache<string,timestamp_t> AccessDeniedCache(4096);
+
 HTTPHeaders CreateS3Header(string url, string query, string host, string service, string method,
                            const S3AuthParams &auth_params, string date_now, string datetime_now, string payload_hash,
-                           string content_type, string content_md5) {
+                           string content_type, string content_md5, string account_id) {
 
 	HTTPHeaders res;
 	res["Host"] = host;
@@ -71,6 +85,10 @@ HTTPHeaders CreateS3Header(string url, string query, string host, string service
 		res["x-amz-request-payer"] = "requester";
 	}
 
+	if(!account_id.empty()) {
+		res["x-amz-account-id"] = account_id;
+	}
+
 	string signed_headers = "";
 	hash_bytes canonical_request_hash;
 	hash_str canonical_request_hash_str;
@@ -84,7 +102,11 @@ HTTPHeaders CreateS3Header(string url, string query, string host, string service
 			res["content-type"] = content_type;
 		}
 	}
-	signed_headers += "host;x-amz-content-sha256;x-amz-date";
+	signed_headers += "host";
+	if (!account_id.empty()) {
+		signed_headers += ";x-amz-account-id";
+	}
+	signed_headers += ";x-amz-content-sha256;x-amz-date";
 	if (use_requester_pays) {
 		signed_headers += ";x-amz-request-payer";
 	}
@@ -101,7 +123,11 @@ HTTPHeaders CreateS3Header(string url, string query, string host, string service
 	if (content_type.length() > 0) {
 		canonical_request += "\ncontent-type:" + content_type;
 	}
-	canonical_request += "\nhost:" + host + "\nx-amz-content-sha256:" + payload_hash + "\nx-amz-date:" + datetime_now;
+	canonical_request += "\nhost:" + host;
+	if (!account_id.empty()) {
+		canonical_request += "\nx-amz-account-id:" + account_id;
+	}
+	canonical_request += "\nx-amz-content-sha256:" + payload_hash + "\nx-amz-date:" + datetime_now;
 	if (use_requester_pays) {
 		canonical_request += "\nx-amz-request-payer:requester";
 	}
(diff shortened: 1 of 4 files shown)


romangershgorin:
@Mytherin sorry to tag you directly, but as an active contributor to this repository, could you please help with review of this MR?

guillesd:
@alisoleimani could you provide an example of how this would be used?

ivanovro:
Thanks for the followup @guillesd!

> @alisoleimani could you provide an example of how this would be used?

As we’re (@romangershgorin & @alisoleimani) part of the same organisation and share an interest in making this capability available, I can add some context on the use cases behind it.

Standard AWS IAM policies have practical limits on the number of statements that can be attached to a role. In large organisations, where access is managed at a fine granularity (for example, per bucket prefix or even per table mapped to S3 paths), these limits quickly become a bottleneck for scalable data access management.

S3 Access Grants address this by providing short-lived, vended credentials that are dynamically issued for specific bucket prefixes. This allows access to be managed centrally and independently of IAM role definitions, avoiding the need to constantly update or expand IAM policies as access requirements evolve.

romangershgorin:
@alisoleimani should it be also added [here](https://github.com/alisoleimani/duckdb-httpfs/blob/main/src/create_secret_functions.cpp#L111) to allow enabling access grants for secret? I had to add it [here](https://github.com/duckdb/duckdb-aws/blob/main/src/aws_secret.cpp#L359) in duckdb-aws extension also, to allow usage like this
```
CREATE SECRET secret (
        TYPE s3,
        PROVIDER credential_chain,
        CHAIN process,
        S3_ACCESS_GRANTS_ENABLED true
);
```

alisoleimani:
@guillesd Roman already described the use case. 

@romangershgorin  I rebased the code to the latest main so it must be buildable now.

To use this feature you first need to enable it, either from attributes or environment var.

```
SET s3_access_grants_enabled=1;
```

```
export DUCKDB_S3_ACCESS_GRANTS_ENABLED=1
```

Now for every read/write to S3, we first check if there is a valid s3 access grant fo that s3 prefix using your identity or not, if your identity has access to that prefix using access grant we update the credentials to what access grants returns, otherwise we use your current credentials.

The core idea is to have the same functionality as  https://github.com/aws/boto3-s3-access-grants-plugin

alisoleimani:
example from our datalake: 
Screenshot 2026-02-02 at 17 29 21

alisoleimani:
> @alisoleimani should it be also added [here](https://github.com/alisoleimani/duckdb-httpfs/blob/main/src/create_secret_functions.cpp#L111) to allow enabling access grants for secret? I had to add it [here](https://github.com/duckdb/duckdb-aws/blob/main/src/aws_secret.cpp#L359) in duckdb-aws extension also, to allow usage like this
> 
> ```
> CREATE SECRET secret (
>         TYPE s3,
>         PROVIDER credential_chain,
>         CHAIN process,
>         S3_ACCESS_GRANTS_ENABLED true
> );
> ```

Depends if we want to add it or not, I do not see a use case for that yet, however if duckdb team asks us to add it to secret we can follow your suggestion

src/s3fs.cpp
```diff
@@ -138,6 +163,193 @@ HTTPHeaders CreateS3Header(string url, string query, string host, string service
...
 
+// Forward declaration for FindTagContents (defined later in file)
+optional_idx FindTagContents(const string &response, const string &tag, idx_t cur_pos, string &result);
+
+string get_current_account_id(HTTPParams &http_params, S3AuthParams &auth_params) {
+	string cached_account_id;
+	if (AccountIdCache.Get(auth_params.access_key_id, cached_account_id)) {
+		return cached_account_id;
+	}
+	string query = "Action=GetCallerIdentity&Version=2011-06-15";
```
Tmonster, line 174: where does `2011-06-15` come from?

src/s3fs.cpp
```diff
@@ -138,6 +163,193 @@ HTTPHeaders CreateS3Header(string url, string query, string host, string service
...
+	if (result->HasRequestError()) {
+		throw IOException("%s error for HTTP GET to '%s'", result->GetRequestError(), full_url);
+	}
+	string account_id;
+	FindTagContents(response.str(), "Account", 0, account_id);
+	AccountIdCache.Put(auth_params.access_key_id, account_id);
+	return account_id;
+}
+
+string get_account_id_for_s3_object(HTTPParams &http_params, S3AuthParams &auth_params, const string &url) {
```
Tmonster: Nit: We support `CamelCase` for function names

src/s3fs.cpp
```diff
@@ -138,6 +163,193 @@ HTTPHeaders CreateS3Header(string url, string query, string host, string service
...
+	    [&](const_data_ptr_t data, idx_t data_length) {
+		    response << string(const_char_ptr_cast(data), data_length);
+		    return true;
+	    });
+	auto result = http_params.http_util.Request(get_data_access_creds);
+	// We ignore all errors here as we want to fallback to normal IAM creds
+	if (result->HasRequestError() || (int)result->status >= 400) {
+		// cache access denied for 5 min
+		if ((int)result->status == 403) {
+			AccessDeniedCache.Put(url_fixed_prefix, Timestamp::GetCurrentTimestamp() + 5 * 60 * 100000);
```
Tmonster: DuckDB timestamps are in microseconds, I think you are missing a 0?

src/s3fs.cpp
```diff
@@ -29,9 +30,21 @@
...
 namespace duckdb {
 
+struct TemporaryAWSCredential {
+	string access_key_id;
+	string secret_access_key;
+	string session_token;
+	timestamp_t expiration {};
+};
+
+static LRUCache<string, string> AccountIdCache(1024);
```
Tmonster, line 40: I think these will be available across all DuckDB connections. So if multiple connections are using  s3 access grants and secrets, it's possible they may pollute the caches, don't know how desireable that is

src/s3fs.cpp
```diff
@@ -138,6 +163,193 @@ HTTPHeaders CreateS3Header(string url, string query, string host, string service
...
+		    return true;
+	    });
+	auto result = http_params.http_util.Request(get_access_grant_urn);
+	if (result->HasRequestError()) {
+		throw IOException("%s error for HTTP GET to '%s'", result->GetRequestError(), full_url);
+	}
+	string access_grants_arn;
+	FindTagContents(response.str(), "AccessGrantsInstanceArn", 0, access_grants_arn);
+	vector<string> parts = StringUtil::Split(access_grants_arn, ':');
+	BucketOwnerAccountIdCache.Put(parsed_url.bucket, parts[4]);
```
Tmonster, line 240: Can we check parts length before accessing? If the tag doesn't exist in the response, this could error

src/s3fs.cpp
```diff
@@ -138,6 +163,193 @@ HTTPHeaders CreateS3Header(string url, string query, string host, string service
...
+			}
+			AccessGrantsCache.Delete(prefix);
+		}
+		current_pos--;
+	}
+	string query = "durationSeconds=3600&permission=" + operation +
+	               "&privilege=Default&target=" + StringUtil::URLEncode(url_fixed_prefix) + "&targetType=Object";
+	string access_grants_url = "/v20180820/accessgrantsinstance/dataaccess";
+	string host = account_id + ".s3-control." + auth_params.region + ".amazonaws.com";
+	auto full_url = "https://" + host + access_grants_url + "/" + query;
```
Tmonster: should we use `?` here to define the query part of the url?

Tmonster (commented):
I need to take a longer look at this and I've left a couple of comments. 

My main thought, however, is that this seems very aws specific, is there a reason you did not implement this in the duckdb-aws extension? Possibly the aws-cpp-sdk has a number of optimizations for this workflow?

src/s3fs.cpp
alisoleimani: you are right, removed this line as it was not used

alisoleimani:
> I need to take a longer look at this and I've left a couple of comments.
> 
> My main thought, however, is that this seems very aws specific, is there a reason you did not implement this in the duckdb-aws extension? Possibly the aws-cpp-sdk has a number of optimizations for this workflow?

As I wanted this flow to happen in s3 access I needed it in s3fs, if there is another possibility to have the same functionality I am open to consider it

src/s3fs.cpp
alisoleimani: Good point, how can I create cache instances per connection?

PR #165, opened 2025-11-14
Add credentials refresh in more locations

J-Meyers:
This is very difficult to test since you must wait for the credentials to timeout, however there were a variety of locations and scenarios that previously did not automatically refresh the credentials.

After the first `S3FileHandle` was `Initialize`d then many operations like GLOBs would never have an opportunity to refresh the handle.

This adds in autorefresh around all calls, the core of this is the `ExecuteWithRefresh` and keeping around the `FileOpener`. `ExecuteWithRefresh` is used to just wrap the old code exactly as it was with a mechanism to automatically refresh the credentials and then run it again.

The previous refreshing in `S3FileHandle::Initialize` is now removed since it is now managed at each of the lower level function calls.

Should close:
https://github.com/duckdb/duckdb-aws/issues/26
https://github.com/duckdb/duckdb-aws/issues/64
https://github.com/duckdb/duckdb-aws/issues/97

If you have any advice on how to create an automated test for this I'd appreciate it

### DIFF
--- a/src/include/s3fs.hpp
+++ b/src/include/s3fs.hpp
@@ -33,7 +33,15 @@ struct S3AuthParams {
 	bool requester_pays = false;
 	string oauth2_bearer_token; // OAuth2 bearer token for GCS
 
+	// Store FileOpener and path for credential refresh
+	optional_ptr<FileOpener> opener;
+	string path;
+
 	static S3AuthParams ReadFrom(optional_ptr<FileOpener> opener, FileOpenerInfo &info);
+
+	//! Try to refresh credentials if they've expired
+	//! Returns true if refresh succeeded and credentials were updated
+	bool TryRefreshCredentials();
 };
 
 struct AWSEnvironmentCredentialsProvider {
@@ -236,6 +244,9 @@ public:
 	static string GetGCSAuthError(S3AuthParams &s3_auth_params);
 	static HTTPException GetS3Error(S3AuthParams &s3_auth_params, const HTTPResponse &response, const string &url);
 
+	//! Helper method to attempt secret refresh for a given path
+	static bool TryRefreshSecret(const string &path, optional_ptr<FileOpener> opener);
+
 protected:
 	static void NotifyUploadsInProgress(S3FileHandle &file_handle);
 	duckdb::unique_ptr<HTTPFileHandle> CreateHandle(const OpenFileInfo &file, FileOpenFlags flags,
@@ -245,6 +256,24 @@ protected:
 	string GetPayloadHash(char *buffer, idx_t buffer_len);
 
 	HTTPException GetHTTPError(FileHandle &, const HTTPResponse &response, const string &url) override;
+
+private:
+	template <typename RequestFunc>
+	auto ExecuteWithRefresh(S3AuthParams &auth_params, RequestFunc request_func)
+	    -> decltype(request_func(auth_params)) {
+		try {
+			return request_func(auth_params);
+		} catch (std::exception &ex) {
+			ErrorData error(ex);
+			// Only attempt refresh for HTTP or IO exceptions (same logic as S3FileHandle::Initialize)
+			if (error.Type() == ExceptionType::IO || error.Type() == ExceptionType::HTTP) {
+				if (auth_params.TryRefreshCredentials()) {
+					return request_func(auth_params); // Retry
+				}
+			}
+			throw;
+		}
+	}
 };
 
 // Helper class to do s3 ListObjectV2 api call https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -228,7 +228,36 @@ S3AuthParams S3AuthParams::ReadFrom(optional_ptr<FileOpener> opener, FileOpenerI
 		result.endpoint = "s3.amazonaws.com";
 	}
 
+	// Store opener and path for potential credential refresh
+	result.opener = opener;
+	result.path = info.file_path;
+
 	return result;
+}
+
+bool S3AuthParams::TryRefreshCredentials() {
+	if (!opener) {
+		return false;
+	}
+
+	// Try to refresh the secret using S3FileSystem helper
+	if (S3FileSystem::TryRefreshSecret(path, opener)) {
+		// Refresh succeeded, reload credentials
+		FileOpenerInfo info = {path};
+		auto refreshed_params = S3AuthParams::ReadFrom(opener, info);
+
+		// Update this object's credentials
+		this->access_key_id = refreshed_params.access_key_id;
+		this->secret_access_key = refreshed_params.secret_access_key;
+		this->session_token = refreshed_params.session_token;
+		this->region = refreshed_params.region;
+		this->endpoint = refreshed_params.endpoint;
+		this->oauth2_bearer_token = refreshed_params.oauth2_bearer_token;
+
+		return true;
+	}
+
+	return false;
 }
 
 unique_ptr<KeyValueSecret> CreateSecret(vector<string> &prefix_paths_p, string &type, string &provider, string &name,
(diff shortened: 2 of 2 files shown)


carlopi:
@J-Meyers: Thanks for the PR, just an heads up that it might take a bit for this to get properly through review process, but we'll get to it.

Testing is indeed not trivial, unsure what's the way forward, but I'll think of something

szalai1:
When can we expect this to be merged? How can I help to test this? @carlopi

mst:
This also bites DuckLake's ability to auto-refresh. The chain:

1. DuckLake passes `file_size` + dummy `etag` and `last_modified` in `extended_info`: https://github.com/duckdb/ducklake/blob/c80b9155c8ff260778b6a5bc726e50d65edc84dc/src/storage/ducklake_multi_file_list.cpp#L195-L198
	```cpp
			extended_info->options["validate_external_file_cache"] = Value::BOOLEAN(false);
			// etag / last modified time can be set to dummy values
			extended_info->options["etag"] = Value("");
			extended_info->options["last_modified"] = Value::TIMESTAMP(timestamp_t(0));
	```
	this seems to be on purpose to avoid unnecessary HEAD requests (https://github.com/duckdb/ducklake/blob/c80b9155c8ff260778b6a5bc726e50d65edc84dc/src/storage/ducklake_delete_filter.cpp#L23)

2. this makes `HTTPFileHandle` ctor flip `initialized = true` ....
https://github.com/duckdb/duckdb-httpfs/blob/13e18b3c9f3810334f5972b76a3acc247b28e537/src/httpfs.cpp#L400-L404

3. ...and makes L829 early-return from `LoadFileInfo()` https://github.com/duckdb/duckdb-httpfs/blob/13e18b3c9f3810334f5972b76a3acc247b28e537/src/httpfs.cpp#L827-L840

**Why this matters**: today's only auto-refresh hook lives in `S3FileHandle::Initialize`'s catch block -- it fires when the **HEAD** returns 403. By design, DuckLake (and Iceberg) skip the HEAD via prefilled `extended_info` to avoid an extra round-trip per file. That skip silently opts the whole read out of auto-refresh: `Initialize()` completes against stale creds, and the first auth check happens in a range read, which has no equivalent hook. STS expiry mid-session then propagates raw `ExpiredToken` to the user, even with `refresh='auto'` configured on the secret.

### Reproduction

```sql
INSTALL postgres; LOAD postgres;
INSTALL ducklake; LOAD ducklake;
FORCE INSTALL aws FROM core_nightly; LOAD aws;
INSTALL httpfs; LOAD httpfs;

CREATE SECRET aws_s3 (
    TYPE S3, PROVIDER credential_chain, CHAIN 'web_identity',
    ASSUME_ROLE_ARN '<role>',
    WEB_IDENTITY_TOKEN_FILE '<jwt-file>',
    REGION 'eu-west-1'
);
ATTACH 'ducklake:postgres:host=<pg> port=5432 dbname=<db>' AS x;
```

| Step | Action | Result | secret `key_id` |
|---|---|---|---|
| 1 | `SELECT * FROM x.main.t1 LIMIT 1` | OK | initial |
| 2 | `time.sleep(65 * 60)` (past STS expiry) | -- | initial (unchanged) |
| 3 | `SELECT * FROM x.main.t2 LIMIT 1` | FAIL `ExpiredToken` | initial |
| 4 | `SELECT * FROM read_parquet('s3://...')` | OK; fires existing `Initialize` refresh | NEW |
| 5 | `SELECT * FROM x.main.t3 LIMIT 1` | OK | NEW (from step 4) |

--- 

EDIT: same pattern on duckdb-iceberg: e.g. https://github.com/duckdb/duckdb-iceberg/blob/b0d93913629a08b91af90da9ae7e89d846cd22ec/src/planning/iceberg_multi_file_list.cpp#L1127-L1132. 
EDIT 2: add more explanation why early-return from `LoadFileInfo()` is a problem

mst:
[refresh_get_range_test.sh](https://github.com/user-attachments/files/27471402/refresh_get_range_test.sh)

This is a small shell-driver integration test that exercises the `GetRangeRequest` path specifically -- i.e. the gap DuckLake / Iceberg hit via prefilled `extended_info`. Reproduces without either of those extensions by using `enable_http_metadata_cache=true` to trigger the same HEAD-skip on the second read.

How the test works: Two service accounts under one parent user. Phase A reads with svcacct A (populates `http_metadata_cache`), then `mc admin user svcacct rm` revokes it. Phase B's first I/O is a range GET (HEAD skipped) signed with the now-revoked token. With this PR the wrapper catches the 403 and refreshes via `REFRESH_INFO` to svcacct B; without it, the 403 propagates.

## How to run

```bash
# 1. minio -- same image pinned in `scripts/minio_s3.yml`
#    (what `scripts/run_s3_test_server.sh` boots in CI via docker compose).
docker run -d --name duckdb-minio -p 9000:9000 \
    -e MINIO_ROOT_USER=duckdb_minio_admin \
    -e MINIO_ROOT_PASSWORD=duckdb_minio_admin_password \
    minio/minio:RELEASE.2021-11-03T03-36-36Z server /data
until curl -sf http://localhost:9000/minio/health/live >/dev/null; do sleep 1; done

MC="docker run --rm --network=host \
    -e MC_HOST_local=http://duckdb_minio_admin:duckdb_minio_admin_password@localhost:9000 \
    minio/mc:RELEASE.2021-11-05T10-05-06Z"
$MC admin user add local minio_duckdb_user minio_duckdb_user_password
$MC admin policy set local readwrite user=minio_duckdb_user
$MC mb local/test-bucket

# 2. build + run the test
make release
ENDPOINT=localhost:9000 URL_STYLE=path \
  DUCKDB=./build/release/duckdb \
  ./scripts/refresh_get_range_test.sh
```

I hope that helps moving this PR forward.

src/include/s3fs.hpp
```diff
@@ -245,6 +256,24 @@ class S3FileSystem : public HTTPFileSystem {
...
+	template <typename RequestFunc>
+	auto ExecuteWithRefresh(S3AuthParams &auth_params, RequestFunc request_func)
+	    -> decltype(request_func(auth_params)) {
+		try {
+			return request_func(auth_params);
+		} catch (std::exception &ex) {
+			ErrorData error(ex);
+			// Only attempt refresh for HTTP or IO exceptions (same logic as S3FileHandle::Initialize)
+			if (error.Type() == ExceptionType::IO || error.Type() == ExceptionType::HTTP) {
+				if (auth_params.TryRefreshCredentials()) {
```
mst, line 270: thread safety: n threads expiring simultaneously will probably each issue their own STS call. This might require a mutex around the Credentials Refresh call and some way to detect that a refresh already happened.

src/s3fs.cpp
```diff
@@ -228,9 +228,38 @@ S3AuthParams S3AuthParams::ReadFrom(optional_ptr<FileOpener> opener, FileOpenerI
...
+		FileOpenerInfo info = {path};
+		auto refreshed_params = S3AuthParams::ReadFrom(opener, info);
+
+		// Update this object's credentials
+		this->access_key_id = refreshed_params.access_key_id;
+		this->secret_access_key = refreshed_params.secret_access_key;
+		this->session_token = refreshed_params.session_token;
+		this->region = refreshed_params.region;
+		this->endpoint = refreshed_params.endpoint;
+		this->oauth2_bearer_token = refreshed_params.oauth2_bearer_token;
```
mst, line 255: this might cause a data race -- a reading thread accessing e.g. the updated `access_key_id` before the `secret_access_key` is being written by the refreshing thread.

Might not be a big functional problem if there's the mutex around the refresh: either the reader runs with an expired token into the refresh mutex, or it does with a invalid, partially updated set of credentials and runs into the refresh mutex (assuming the exception handling for expired tokens == invalid signature). 

Not sure if this is a problem, but could be solved by swapping all credential attributes atomically.

src/include/s3fs.hpp
```diff
@@ -245,6 +256,24 @@ class S3FileSystem : public HTTPFileSystem {
...
 
 	HTTPException GetHTTPError(FileHandle &, const HTTPResponse &response, const string &url) override;
+
+private:
+	template <typename RequestFunc>
+	auto ExecuteWithRefresh(S3AuthParams &auth_params, RequestFunc request_func)
+	    -> decltype(request_func(auth_params)) {
+		try {
+			return request_func(auth_params);
+		} catch (std::exception &ex) {
```
mst, line 266: It looks like `HTTPFileSystem::PostRequest`  doesn't throw on 4xx [1] (and neither does Put/Delete/Head [2] [3]), so this also needs a check on 4xx. Only GetRequest does [4], and that's why the test I posted succeded...

[1] https://github.com/duckdb/duckdb-httpfs/blob/main/src/httpfs.cpp#L164 
[2] https://github.com/duckdb/duckdb-httpfs/blob/main/src/httpfs.cpp#L167
[3] https://github.com/duckdb/duckdb-httpfs/blob/main/src/httpfs.cpp#L193
[4] https://github.com/duckdb/duckdb-httpfs/blob/main/src/httpfs.cpp#L242

bmulh:
I am running into a issue that seems related to this.  At exactly an hour after creating the secret I get:
```
  HTTP Error: HTTP GET error reading 's3://<s3_path>' in region 'us-east-1' (HTTP 400 Bad Request)
  
  ExpiredToken: The provided token has expired.
```
We are running dbt with duckdb(v1.5.3) and one of the tables is written to s3 using a secret with `web_identity`.

Is there any guidance on a short term solution or when this may be merged? Thanks!

kinghuang:
This PR has been superseded by #332. I'm hoping the latter will be merged before the fall DuckDB 2.0 release!

lnkuiper:
This PR has been superseded by work on v1.5 and main

PR #167, opened 2025-11-17
Caching HTTPClients across queries

carlopi:
This PR proposes to add a new HTTPUtils, modelled after HTTPFSUtils, that allow for caching of `HTTPFSCachedClient`s across queries.

This has the somewhat visible advantage that reduces the time spent setting up openssl secure context and validating certificates on the resource we are connecting to.

What it's tested in CI for this PR is that no behavioural changes (on the tests) are there if one keep using the default backend (that is `httplib`) AND that code works also when setting on init `httplib-cached` (new, added via IntegrationTests.yml).

Demo:
```sql
LOAD httpfs;
.timer on
SELECT count(*) FROM 's3://overturemaps-us-west-2/release/2025-10-22.0/theme=base/type=infrastructure/*.parquet';
SELECT count(*) FROM 's3://overturemaps-us-west-2/release/2025-10-22.0/theme=base/type=infrastructure/*.parquet';
```
Might time at around 4.3 seconds for the first query (cold) and then 0.6/0.7s for subsequent ones
```sql
LOAD httpfs;
SET httpfs_client_implementation = 'httplib-cached';
.timer on
SELECT count(*) FROM 's3://overturemaps-us-west-2/release/2025-10-22.0/theme=base/type=infrastructure/*.parquet';
SELECT count(*) FROM 's3://overturemaps-us-west-2/release/2025-10-22.0/theme=base/type=infrastructure/*.parquet';
SET httpfs_client_implementation = 'httplib-cached';
SELECT count(*) FROM 's3://overturemaps-us-west-2/release/2025-10-22.0/theme=base/type=infrastructure/*.parquet';
```
Might time at around 3.8 seconds for the first query (cold) and then 0.2/0.3s for subsequent ones

Note that this is very significant when caching is performed cross queries (0.6/0.7 -> 0.2/0.3), but also somewhat visible also within the first same query.

Doing:
```sql
SET httpfs_client_implementation = 'httplib-cached';
```
will as side effect clean-up the cache.

Some notes:
* to make this more clearly opt-in, code is duplicated. Given importance of httpfs, I think this was the best path, so that currently executed code-paths are more provably be exactly the same. This is an implementation detail, the state up to https://github.com/duckdb/duckdb-httpfs/commit/061a03a04d2fb34f3e667a0f225263f4384bb73d has this on top of the existing  HTTPFUtil
* `clients` are in some case overridden, and there is no clear way (at the moment) to avoid this behaviour (iff you opt in cached clients, that is)
* try_lock is using for cache accesses and updating. Meaning that we'll never wait for cache contention. This might not be optimal, but ensure no slowdowns
* no access time is stored in the cache, either we find a spot where to save a HTTPClient, or we pick a spot at random to vacate. This simplifies reasoning, but maybe explicit time management would be better
* this rework hit a few areas that were weird in the code, in particular `SetHttpfsClientImplementation` being unused and the note at https://github.com/duckdb/duckdb-httpfs/compare/v1.4-andium...carlopi:duckdb_httpfs:caching_clients?expand=1#diff-a89bf9b2e675b51de8853bbd21c29dc9adbe7e1cdc7109ecbae09d1be1f10a7aR102), were current code is outdated, and needs either to be removed or adapted

### DIFF
--- a/.github/workflows/IntegrationTests.yml
+++ b/.github/workflows/IntegrationTests.yml
@@ -86,3 +86,4 @@ jobs:
           source ./scripts/run_s3_test_server.sh
           source ./scripts/set_s3_test_server_variables.sh
           ./build/release/test/unittest "*" --skip-error-messages "[]"
+          ./build/release/test/unittest "*" --skip-error-messages "[]" --on-init "SET httpfs_client_implementation = 'httplib-cached';" --statically-loaded-extensions "[core_functions,parquet,httpfs]"
--- a/.github/workflows/IntegrationTests.yml
+++ b/.github/workflows/IntegrationTests.yml
@@ -86,4 +86,5 @@ jobs:
           source ./scripts/run_s3_test_server.sh
           source ./scripts/set_s3_test_server_variables.sh
           ./build/release/test/unittest "*" --skip-error-messages "[]"
-          ./build/release/test/unittest "*" --skip-error-messages "[]" --on-init "SET httpfs_client_implementation = 'httplib-cached';" --statically-loaded-extensions "[core_functions,parquet,httpfs]"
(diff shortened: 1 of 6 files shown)


carlopi:
2 test failures, one random (Windows) the other unclear how to work around, but will figure it out (problem is that a file is left around).

PR #171, opened 2025-12-02
Fix HTTP header merging

samansmink:
This PR is a first step to change duckdb's behaviour for HTTP secrets used in s3 requests.

### Main problem solved
Currently, `HTTP` secrets **are** used for the http requests performed for S3 requests. E.g.:

```SQL
CREATE SECRET (
    TYPE HTTP,
    EXTRA_HTTP_HEADERS MAP{
		'CustomHeader': 'fliepflap'
	}
);
FROM 's3://test-bucket/http-secret-test/test.parquet'
```

In this case, the request to `s3://test-bucket/http-secret-test/test.parquet` will have the `CustomHeader` value attached. 

This is kindof confusing, since users might not be aware that the HTTP secrets are behaving like this. This PR kicks of the process to change this behaviour by adding a (for now default `true`) setting `merge_http_secret_into_s3_request`. This setting now changes the behaviour and no longer merges the http secret into the s3 requests.

My idea is to flip the default of this setting for duckdb `v1.5` to `false`. Then there will be a workaround available for those relying on this behaviour.

Now disabling this prevents users from being able to set http proxy settings and extra_header settings for s3 requests. Therefore i simply copied over these settings to be set in the S3 secret.  This allows:

```SQL
CREATE SECRET (
    TYPE S3,
    PROVIDER config,
    KEY_ID '...',
    SECRET '...',
    EXTRA_HTTP_HEADERS MAP{
    		'CustomHeader': 'fliepflap'
    	}
)
FROM 's3://test-bucket/http-secret-test/test.parquet'
```

### Side quest
When writing tests i noticed that the headers from the `EXTRA_HTTP_HEADERS` field were not visible in the logs. I made a small rewrite to ensure these headers are now properly visible in the logs (and this should also correctly forward them to the non httlib HTTPUtil requests).

This is unfortunately not yet fixed in all cases, but I've opened a draft PR to DuckDB https://github.com/duckdb/duckdb/pull/20030 to fix this everywhere.


carlopi:
Current behaviour though will change, from scanning `http` only to scanning all 5 secret types, it's that on purpose / necessary?

carlopi:
I think now looks better, but there is still something that might need clean-up in duckdb: moving logic out of `TransformHeaders` and into some callsite (GetRequest / PostRequest & co) I am not sure it's correct, since in the general form callers are not guaranteed to pass from those requests. Think for example extension install, code is like:
```c++
        GetRequestInfo get_request(url, headers, *params, nullptr, nullptr);
        get_request.try_request = true;

        auto response = http_util.Request(get_request); 
```
where `TranformHeaders` would be called from Request, but GetRequest would not be invoked.

Maybe leaving the `TranformHeaders` the same could work? Unsure there.

samansmink:
Thanks for the extensive review @carlopi! applied the requested changes. Could use another set of eyes on this now!

src/httpfs.cpp
```diff
@@ -65,23 +67,42 @@ unique_ptr<HTTPParams> HTTPFSUtil::InitializeParameters(optional_ptr<FileOpener>
 		}
 	}
 
+	unique_ptr<KeyValueSecretReader> settings_reader;
+	string s3_prefix = S3FileSystem::TryGetPrefix(info->file_path);
+	if (info && !s3_prefix.empty()) {
+		// This is an S3-type url, we should
```
carlopi: I think `S3FileSystem::TryGetPrefix` needs also to be guarded by `if (info)`

src/httpfs.cpp
```diff
@@ -65,23 +67,42 @@ unique_ptr<HTTPParams> HTTPFSUtil::InitializeParameters(optional_ptr<FileOpener>
...
+		const char *s3_secret_types[] = {"s3", "r2", "gcs", "aws", "http"};
```
carlopi, line 73: Should we not use only the right token then? Like s3:// only when path is s3:// ?

carlopi (commented):
More comments, but looks almost right

src/httpfs.cpp
samansmink: oops, thanks

src/httpfs.cpp
samansmink: no, secret scope takes care of that!

carlopi:
Thanks!

PR #173, opened 2025-12-09
Expose `create_s3_header` and add new API for `S3AuthParams::ReadFrom`

NiclasHaderer:
This exposes a previously internal function (`create_s3_header`) and splits up `S3AuthParams::ReadFrom` to allow creating `S3AuthParams` from a `KeyValueSecretReader`


NiclasHaderer:
@Mytherin does this look good to you, or should I change something? Thanks!

carlopi (commented):
I personally think this is neat, I would also considering adding (for duckdb's `v1.5-variegata` branch / duckdb-httpfs's `main` branch) this to HTTPUtil, probably changing `S3AuthParams` to a `map<string,string>`.

Idea would be that then this can be used (once `httpfs` has been loaded) anywhere.

Mytherin:
Thanks!

PR #174, opened 2025-12-09
Logarithmically grow read buffer for sequential reads

lnkuiper:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/6807#issuecomment-3631452424

Just some basic functionality, but it could be improved using stats in the future.

### DIFF
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -433,14 +433,47 @@ unique_ptr<FileHandle> HTTPFileSystem::OpenFileExtended(const OpenFileInfo &file
 	return std::move(handle);
 }
 
+void HTTPFileHandle::AddStatistics(idx_t read_offset, idx_t read_length, idx_t read_duration) {
+	range_request_statistics.push_back({read_offset, read_length, read_duration});
+}
+
+void HTTPFileHandle::AdaptReadBufferSize(idx_t next_read_offset) {
+	if (range_request_statistics.empty()) {
+		return; // No requests yet - nothing to do
+	}
+
+	const auto &last_read = range_request_statistics.back();
+	if (last_read.offset + last_read.length != next_read_offset) {
+		return; // Not reading sequentially
+	}
+
+	if (read_buffer.GetSize() >= MAXIMUM_READ_BUFFER_LEN) {
+		return; // Already at maximum size
+	}
+
+	// Grow the buffer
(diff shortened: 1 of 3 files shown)


src/httpfs.cpp
```diff
@@ -842,7 +884,7 @@ void HTTPFileHandle::Initialize(optional_ptr<FileOpener> opener) {
 				etag = value.etag;
 
 				if (flags.OpenForReading()) {
-					read_buffer = duckdb::unique_ptr<data_t[]>(new data_t[READ_BUFFER_LEN]);
+					read_buffer = AllocateReadBuffer(opener);
```
carlopi: I think this (possibly in a follow up) can be completely skipped for `flags.DirectIO() ||flags.RequireParallelAccess()` (both here and below) since then we will skip the buffer.

That should avoid a bunch of allocations, that are likely not the bottleneck, but should just be better.

Or maybe better: `AllocateReadBuffer` should take in consideration the flags, and potentially avoid allocating at all

lnkuiper: Good idea. Added this to the PR

carlopi (commented):
This looks great.

I have not tested this locally yet. Would it make sense to add a test? Say something like:
```
statement ok
CALL enable_logging('HTTP');

--- or from elsewhere in the internet
statement ok
FROM read_json('https://raw.githubusercontent.com/duckdb/duckdb/main/data/...');

query I
SELECT count(*) FROM duckdb_logs_parsed('HTTP');
----
7?
```

carlopi:
I was looking for such a stable and publicly hosted file, for example:
```sql
CALL enable_logging('HTTP');
CREATE TABLE T AS FROM 'https://data.gharchive.org/2023-02-08-0.json.gz';
SELECT count(*), request.type FROM duckdb_logs_parsed('HTTP') GROUP BY request.type;
```
is currently:
```
┌──────────────┬─────────┐
│ count_star() │  type   │
│    int64     │ varchar │
├──────────────┼─────────┤
│            1 │ HEAD    │
│          124 │ GET     │
└──────────────┴─────────┘
```
and with this PR becomes:
```
┌──────────────┬─────────┐
│ count_star() │  type   │
│    int64     │ varchar │
├──────────────┼─────────┤
│            1 │ HEAD    │
│            9 │ GET     │
└──────────────┴─────────┘
```

carlopi:
Thanks!!

PR #181, opened 2025-12-28
Implement RemoveFiles for S3FileSystem

utay:
Following [this discussion](https://github.com/duckdb/ducklake/discussions/384) and https://github.com/duckdb/duckdb/pull/20333 which introduces `RemoveFiles` on `FileSystem`, this PR implements `RemoveFiles` for `S3FileSystem`.

It uses the S3 [DeleteObjects](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html) API to delete objects in batches of 1000. 

I couldn't find how to test this, as the other methods (like `RemoveFile`) don't seem to be unit tested, and the `RemoveFiles` method is not exposed as a SQL command. Please let me know if there's a way to do so! I was still able to test it end-to-end with a SQL test in the ducklake repo (PR incoming).

### DIFF
--- a/src/include/s3fs.hpp
+++ b/src/include/s3fs.hpp
@@ -225,6 +225,7 @@ public:
 		return false;
 	}
 	void RemoveFile(const string &filename, optional_ptr<FileOpener> opener = nullptr) override;
+	void RemoveFiles(const vector<string> &filenames, optional_ptr<FileOpener> opener = nullptr) override;
 	void RemoveDirectory(const string &directory, optional_ptr<FileOpener> opener = nullptr) override;
 	void FileSync(FileHandle &handle) override;
 	void Write(FileHandle &handle, void *buffer, int64_t nr_bytes, idx_t location) override;
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1028,15 +1028,17 @@ void S3FileSystem::RemoveFiles(const vector<string> &paths, optional_ptr<FileOpe
 			string_t md5_blob(const_char_ptr_cast(md5_hash), MD5Context::MD5_HASH_LENGTH_BINARY);
 			string content_md5 = Blob::ToBase64(md5_blob);
 
-			string http_query_param = "delete=";
+			const string http_query_param_for_sig = "delete=";
+			const string http_query_param_for_url = "delete";
 			auto payload_hash = GetPayloadHash(const_cast<char *>(body.data()), body.length());
 
-			auto headers = CreateS3Header(url_info.path, http_query_param, url_info.host, "s3", "POST",
-			                              url_info.auth_params, "", "", payload_hash, "application/xml");
+			auto headers = CreateS3Header(url_info.path, http_query_param_for_sig, url_info.host, "s3", "POST",
+			                              url_info.auth_params, "", "", payload_hash, "");
 			headers["Content-MD5"] = content_md5;
 			headers["Content-Type"] = "application/xml";
 
-			string http_url = url_info.http_proto + url_info.host + url_info.path + "?" + http_query_param;
+			string http_url = url_info.http_proto + url_info.host + S3FileSystem::UrlEncode(url_info.path) + "?" +
+			                  http_query_param_for_url;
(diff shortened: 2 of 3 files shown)


pdet:
Hi @utay, can you retarget this PR to `main`? I'm also unaware of a way to test the deletion function directly. Maybe @samansmink or @carlopi has an idea?

utay:
> Hi @utay, can you retarget this PR to `main`?

Done!

> I'm also unaware of a way to test the deletion function directly. Maybe @samansmink or @carlopi has an idea?

Yep, let me know, happy to help write a test

carlopi:
I need to bump duckdb further, since currently it doesn't include https://github.com/duckdb/duckdb/pull/20333 (see failure in MinIO tests). I'll ping once this can be rebased.

carlopi:
@utay: can you merge with / rebase to the `main` branch? Should be ready after that CI side.

I think the change set might slightly change behaviour, and would indeed need to be more properly tested.

I am not sure of what strategy to recommend, probably combination of populating files + globs + deletion + globs in SQL to be execute as integration test would be handy.

utay:
@carlopi Thanks, I've rebased the branch on `main`. 

I’ll fix the compilation issues and try to add an integration test as you suggested when I have some time!

utay:
@carlopi I've fixed compilation, CI should pass now.

Regarding the integration test, I think there are 2 options:
- Rely on the `COPY` overwrite mode, but this would require removing [this check](https://github.com/duckdb/duckdb/blob/main/src/execution/operator/persistent/physical_copy_to_file.cpp#L413-L417)
 in DuckDB. Also, this doesn’t explicitly delete the existing files; it overwrites them implicitly
- Introduce a SQL function like `s3_remove_files` (or similar) to explicitly clean up the target path. There don’t seem to be any existing S3 helper functions for this today, but adding one might be acceptable

I couldn't find another way to test this with the existing test infrastructure, let me know what you think!

carlopi:
@utay: I've opened https://github.com/duckdb/duckdb/pull/20487, I think removing the check is the path forward, but need to check with the team.

utay:
Hi @carlopi, I was able to write a test now that `COPY ... OVERWRITE` works for remote filesystems. I just needed a small follow-up change to get it fully working: https://github.com/duckdb/duckdb/pull/20576.

utay:
I just rebased on `main` to benefit from https://github.com/duckdb/duckdb-httpfs/pull/217, the test and CI should pass now!

carlopi (approved):
Thanks! This looks to be working as expected, great cross-repositories work.

(while testing, I think I found a bug in the implementation of `OVERWRITE_OR_IGNORE`, but that's to be fixed in duckdb/duckdb)

EDIT: `OVERWRITE_OR_IGNORE` it's just weird.

carlopi:
Actually there was an issue with creating `sigv4` signed requests I had not realized, fixed in a follow up at https://github.com/duckdb/duckdb-httpfs/pull/234

utay:
> Actually there was an issue with creating `sigv4` signed requests I had not realized, fixed in a follow up at #234

I didn't realize either. Thank you for following up!

carlopi:
There have been a report in ducklake that this is not compatible with R2 API, possibly due to some interaction with other work.

I'll have a look at reproducing this.

utay:
> There have been a report in ducklake that this is not compatible with R2 API, possibly due to some interaction with other work.
> 
> I'll have a look at reproducing this.

Arg, I saw [it](https://github.com/duckdb/ducklake/issues/766) pass by. At the time, I tested with path-style MinIO locally and it worked well (and this is what the sqllogictest does too). We're using R2 in production so I'm interested in the potential bug fix, let me know if I can help

carlopi:
I have something else up first, then I'll look into this, I think given timeframe maybe it's handy if I can ping you if I have a PR up, and you can have a look yourself?

utay:
> I have something else up first, then I'll look into this, I think given timeframe maybe it's handy if I can ping you if I have a PR up, and you can have a look yourself?

Sure thing!

PR #193, opened 2026-01-09
CURL: Encode spaces

DinosL:
`curl` expects a fully valid, already encoded url when executing a request, whereas`httplib` constructs requests from different components (e.g. host, path) and silently encodes them.

[CURLU_URLENCODE](https://everything.curl.dev/helpers/url/get.html#curlu_urlencode) is not an option here since it only encodes the host part (see manual). I think, it would only be useful if we pass individual components (or split the full url).


carlopi:
Thanks!

PR #196, opened 2026-01-12
Switch default networking backend to curl

DinosL:
Some tests currently fail when using the `curl` backend. These tests are temporarily forced to use `httplib` instead.

`url_encode.test`: should be fixed by https://github.com/duckdb/duckdb-httpfs/pull/193

### DIFF
--- a/src/httpfs_extension.cpp
+++ b/src/httpfs_extension.cpp
@@ -91,13 +91,13 @@ static void LoadInternal(ExtensionLoader &loader) {
 			                            "`default` are currently supported for duckdb-wasm");
 		}
 #ifndef EMSCRIPTEN
-		if (value == "curl") {
+		if (value == "curl" || value == "default") {
 			if (!config.http_util || config.http_util->GetName() != "HTTPFSUtil-Curl") {
 				config.http_util = make_shared_ptr<HTTPFSCurlUtil>();
 			}
 			return;
 		}
-		if (value == "httplib" || value == "default") {
+		if (value == "httplib") {
 			if (!config.http_util || config.http_util->GetName() != "HTTPFSUtil") {
 				config.http_util = make_shared_ptr<HTTPFSUtil>();
 			}
@@ -116,7 +116,7 @@ static void LoadInternal(ExtensionLoader &loader) {
 	if (config.http_util && config.http_util->GetName() == "WasmHTTPUtils") {
 		// Already handled, do not override
 	} else {
-		config.http_util = make_shared_ptr<HTTPFSUtil>();
+		config.http_util = make_shared_ptr<HTTPFSCurlUtil>();
 	}
 
 	auto provider = make_uniq<AWSEnvironmentCredentialsProvider>(config);
--- a/test/sql/copy/s3/download_config.test
+++ b/test/sql/copy/s3/download_config.test
@@ -119,7 +119,7 @@ Unable to connect to URL "http://test-bucket-public.
 statement error
 SELECT i FROM "http://test-bucket-public.duckdb-minio-non-existent-host.com:9000/root-dir/non-existent-file-ljaslkjdas.parquet" LIMIT 3
 ----
-<REGEX>:.*IO ERROR.*HTTP HEAD to 'http://test-bucket-public.duckdb-minio-non-existent-host.*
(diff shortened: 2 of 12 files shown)


src/httpfs_curl_client.cpp
```diff
@@ -406,6 +406,28 @@ class HTTPFSCurlClient : public HTTPClient {
...
+			return "Forbidden";
+		case 404:
+			return "Not Found";
+		case 503:
+			return "ServiceUnavailable_503";
+		default:
+			return "";
+		}
+	}
+
```
carlopi: Could we maybe reuse a combination of:
```
        static HTTPStatusCode ToStatusCode(int32_t status_code);
        static string GetStatusMessage(HTTPStatusCode status);
```
in duckdb/duckdb ?
I guess they should cover most of what you have here, and much more, and we keep a single mapping.

src/s3fs.cpp
```diff
@@ -1237,13 +1237,12 @@ string AWSListObjectV2::Request(string &path, HTTPParams &http_params, S3AuthPar
...
 	}
 
-	string listobjectv2_url = req_path + "?" + req_params;
-
 	auto header_map =
 	    CreateS3Header(req_path, req_params, parsed_url.host, "s3", "GET", s3_auth_params, "", "", "", "");
 
 	// Get requests use fresh connection
 	string full_host = parsed_url.http_proto + parsed_url.host;
+	string listobjectv2_url = full_host + req_path + "?" + req_params;
```
carlopi, line 1244: @DinosL: it's this handled OK in the case of `httplib`?

DinosL: Yes

carlopi:
Locally I got a:
```
GEN=ninja make
./build/release/test/unittest
```
```
1. test/sql/logging/http_logging.test:25
Wrong result in query! (test/sql/logging/http_logging.test:25)!
```

I am not sure what's different between local and CI, I'll give it a look.

I think locally `httplib` I have all passing, while `curl` 1 failure on logging, that's potentially less relevant:
```
./build/release/test/unittest --skip-error-messages "[]" --statically-loaded-extensions "[httpfs,parquet,core_functions]" --on-init "SET httpfs_client_implementation='httplib';"
./build/release/test/unittest --skip-error-messages "[]" --statically-loaded-extensions "[httpfs,parquet,core_functions]" --on-init "SET httpfs_client_implementation='curl';"

**Comment by DinosL** [comment]
Can you show me the full error?

**Comment by carlopi** [comment]
Sure:
```
Wrong result in query! (test/sql/logging/http_logging.test:25)!
SELECT request.headers['Range'], response.headers['Content-Range']
FROM duckdb_logs_parsed('HTTP')
WHERE request.type='GET';
Mismatch on row 1, column response.headers['Content-Range'](index 2)
NULL <> bytes 0-1275/1276
Expected result:
bytes=0-1275	bytes 0-1275/1276

Actual result:
bytes=0-1275	NULL


~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
unittest is a Catch v2.13.7 host application.
Run with -? for options

-------------------------------------------------------------------------------
test/sql/logging/http_logging.test
-------------------------------------------------------------------------------
```

src/httpfs_curl_client.cpp
DinosL: Done!

DinosL:
That's strange! I cannot reproduce locally and CI doesn't seem to fail either.

carlopi:
@DinosL: I rebased this on origin/main, I hope I have not messed up too much your workflow. Idea is checking this + other PR (now landed) where it stands.

carlopi:
After merging other 2 PRs and rebase, this looks great.

Could you review the test modifications, what needs to stay in and what could be reverted?
- [x] test/sql/copy/s3/http_secret.test
- [x] test/sql/copy/s3/url_encode.test
- [x] test/sql/httpfs/hffs.test_slow
- [x] test/sql/secrets/create_secret_gcs.test_slow
- [x] test/sql/full_file_download_fallback.test

I think at least some have been solved now.

DinosL:
Agreed, they should be fixed now. I reverted them and hope CI agrees. 😃

carlopi:
Thanks!!

PR #197, opened 2026-01-12
Integration.yml: Test both curl and httplib explicitly

carlopi:
I am not sure if syntax is proper or if all test handles re-running correctly, but we'll see


carlopi:
Note: this needs to be review manually, since failures are at the moment non visible.

PR #204, opened 2026-01-14
Allow non-standard "GET request with body"

yan-alex:
https://github.com/duckdblabs/duckdb-internal/issues/7011


carlopi:
I am not really sure how the Integration Tests build fails, could you have a look, ideally reproducing that? otherwise we can check together later.

yan-alex:
I'm also a bit puzzled by this @carlopi 
Ill try to re-run it, maybe it was a fluke?

carlopi (commented):
Thanks! Also for fixing the merge commit.

One question I have: but can also be done later, should we actually test that both methods works the same? (And how?)

After this is merged, could you send a PR to duckdb/duckdb@v1.5-variegata removing the patch there and bumping to what will be then the latest commit on the duckdb/duckdb-httpfs@main branch? Thanks

carlopi:
@yan-alex: latest error at:
```
1. test/sql/copy/s3/s3_hive_partition.test:101
Query unexpectedly succeeded! (test/sql/copy/s3/s3_hive_partition.test:101)!
COPY (SELECT * FROM t1) TO 's3://test-bucket/hive-partitioning/filter-test-parquet' (FORMAT PARQUET, PARTITION_BY c, OVERWRITE);
Count	
BIGINT	
[ Rows: 1]
100
```
makes sense, although I am not sure why that's not flagged by DuckDB CI.

This has been changed via https://github.com/duckdb/duckdb/pull/20487, and the test could be adapted to pass, like:
```
statement ok
COPY (SELECT * FROM t1) TO 's3://test-bucket/hive-partitioning/filter-test-parquet' (FORMAT PARQUET, PARTITION_BY c, OVERWRITE);

# FIXME: also test that content of folder is actually what it claims to be
```

Thanks

yan-alex:
> One question I have: but can also be done later, should we actually test that both methods works the same? (And how?)

So I spent quite a while trying to do some good testing. I tried doing it with 
```sql
query I
select request.type FROM duckdb_logs_parsed('HTTP')
----
POST
```
but `duckdb_logs_parsed('HTTP')` was not reliable for getting the actual method sent, since it would keep displaying the old method even when a custom method was set, and verified via curl debug logs. 
I had an idea of then reading the curl debug logs in the sqltests to determine the method being sent, but at the end of the day me and Sam decided it's not worth the effort as testing this is quite tricky. (for curl, no idea if httplib also provides these types of debug logs).
So we settled for manual verification for now.
Also I was testing with code similar to `test/sql/copy/no_head_on_write.test` as that's the only way I found to trigger a `POST`. That test does a multiupload, but the problem was, because the first part of the multiupload (an initial `POST` without body) errored out on the s3 server (expectedly so, because it was a `GET`) the second `POST` with a body never actually got sent, making it even harder to test.

carlopi (approved):

carlopi:
Thanks!

PR #205, opened 2026-01-14
Curl: fix authentication issues

DinosL:
Curl was not using the data from `extra_headers` that contained the authentication information. 

In addition, it was wrongly encoding custom headers to query params. I removed this code entirely, but give me your thoughts about that. No tests were ever reaching this code, so it's unclear to me why it was there.

First commit is cherry-picked from https://github.com/duckdb/duckdb-httpfs/pull/196


carlopi:
Thanks!

PR #206, opened 2026-01-14
Curl: fix pagination problems

DinosL:
Curl was not capturing the response from `GET` requests, resulting in missing pagination metadata (e.g., cursor) required for paginated queries.

built on top of https://github.com/duckdb/duckdb-httpfs/pull/205


src/httpfs_curl_client.cpp
```diff
@@ -222,6 +223,13 @@ class HTTPFSCurlClient : public HTTPClient {
 			state->total_bytes_received += bytes_received;
 		}
 
+		if (info.response_handler) {
+			auto response = TransformResponseCurl(res);
+			if (!info.response_handler(*response)) {
+				return response;
+			}
+		}
+
```
carlopi, line 232: Could you possibly check the similar code in httpfs_httplib_client.cpp:
```c++
        unique_ptr<HTTPResponse> Get(GetRequestInfo &info) override {
                if (state) {
                        state->get_count++;
                }
                auto headers = TransformHeaders(info.headers, info.params);
                if (!info.response_handler && !info.content_handler) {
                        return TransformResult(client->Get(info.path, headers));
                } else {
                        return TransformResult(client->Get(
                            info.path.c_str(), headers,
                            [&](const duckdb_httplib_openssl::Response &response) {
                                    auto http_response = TransformResponse(response);
                                    return info.response_handler(*http_response);
                            },
                            [&](const char *data, size_t data_length) {
                                    if (state) {
                                            state->total_bytes_received += data_length;
                                    }
                                    return info.content_handler(const_data_ptr_cast(data), data_length);
                            }));
                }
        }
```
In particular, I am not sure if they can be made more similar to each other the GET handling.

carlopi:
Thanks!

PR #207, opened 2026-01-16
Apply patches from duckdb/duckdb

maiadegraaf:
This PR applies patches from duckdb/duckdb:
- `fix.patch`


carlopi:
I think this has been already applied, thanks!

PR #209, opened 2026-01-18
Fix file existence for httpfs

### LINKED ISSUE #189: Question: `HTTPFileSystem::FileExists` return false for empty file
Hi team, when reading through the code, I found we decide a file non-existent if its file size is 0.
https://github.com/duckdb/duckdb-httpfs/blob/aea7ac898ed058361c312f714bf4c0fbbf1b5f82/src/httpfs.cpp#L668-L670

Since we don't specify `ReturnNullIfNotExist`, so if the requested file doesn't exist, we should throw an exception and return false here (though here we don't differentiate non-existent and IO exception...).

Thank you!

dentiny:
Closes https://github.com/duckdb/duckdb-httpfs/issues/189

### DIFF
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -664,7 +664,7 @@ string HTTPFileSystem::GetVersionTag(FileHandle &handle) {
 bool HTTPFileSystem::FileExists(const string &filename, optional_ptr<FileOpener> opener) {
 	try {
 		auto handle = OpenFile(filename, FileFlags::FILE_FLAGS_READ, opener);
-		(void) handle; // suppress warning
+		(void)handle; // suppress warning
 		return true;
 	} catch (...) {
 		return false;
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -664,10 +664,7 @@ string HTTPFileSystem::GetVersionTag(FileHandle &handle) {
 bool HTTPFileSystem::FileExists(const string &filename, optional_ptr<FileOpener> opener) {
 	try {
 		auto handle = OpenFile(filename, FileFlags::FILE_FLAGS_READ, opener);
-		auto &sfh = handle->Cast<HTTPFileHandle>();
-		if (sfh.length == 0) {
-			return false;
-		}
+		(void) handle; // suppress warning
 		return true;
 	} catch (...) {
 		return false;


carlopi (commented):
Could you expand on why the change is beneficial?

I guess this is saying length is never non-0, but it's that actually true?

If we are going to touch the code/handle the issue, then I'd say passing ReturnNullIfNotExist to the OpenFile flags and checking return value vs `nullptr` seems better (although, that could change behaviour, and would need some proper testing).

dentiny:
> Could you expand on why the change is beneficial?

Hi Carlo, thanks for the quick reply!
In simple words
- This PR is try to return true for `FileExists` for empty files, which I think is the correct semantics for existence check.
- IMO, as long as an `OpenFile` passes, the file should be decided as "exists", regardless of its file size.

How to reproduce:
- I wrote a python http server to serve local files (code attached below)
- I created an empty file at local filesystem, which will be served by the http server
- I call `FileExists` with http filesystem on the file, which returns false, but I expect it to be true

Create local empty file
```sh
vscode@bf21a2fa9d74$ mkdir /tmp/httpfs_test
vscode@bf21a2fa9d74$ touch /tmp/httpfs_test/empty_file
```

HTTP server impl:
```py
#!/usr/bin/env python3
from http.server import ThreadingHTTPServer, BaseHTTPRequestHandler
from pathlib import Path
from urllib.parse import unquote, urlparse

ROOT = Path("/tmp/httpfs_test").resolve()

class Handler(BaseHTTPRequestHandler):
    def _local_path(self):
        req_path = unquote(urlparse(self.path).path).lstrip("/")
        local_path = (ROOT / req_path).resolve()

        # prevent "../" traversal
        if not str(local_path).startswith(str(ROOT) + "/"):
            return None
        return local_path

    def _send_file_headers(self, local_path: Path):
        size = local_path.stat().st_size
        self.send_response(200)
        self.send_header("Content-Type", "application/octet-stream")
        self.send_header("Content-Length", str(size))
        self.end_headers()

    def do_HEAD(self):
        local_path = self._local_path()
        if local_path is None:
            self.send_response(403); self.end_headers(); return
        if not local_path.is_file():
            self.send_response(404); self.end_headers(); return

        self._send_file_headers(local_path)

    def do_GET(self):
        local_path = self._local_path()
        if local_path is None:
            self.send_response(403); self.end_headers()
            self.wfile.write(b"forbidden\n")
            return
        if not local_path.is_file():
            self.send_response(404); self.end_headers()
            self.wfile.write(b"not found\n")
            return
        self._send_file_headers(local_path)

        # send body
        with open(local_path, "rb") as f:
            self.wfile.write(f.read())

def main():
    host, port = "127.0.0.1", 8000
    print(f"Serving {ROOT} at http://{host}:{port}/<filename>")
    ThreadingHTTPServer((host, port), Handler).serve_forever()

if __name__ == "__main__":
    main()
```
C++ program to check file existence:
```cpp
auto httpfs = make_uniq<HTTPFileSystem>();
auto exists = httpfs->FileExists("http://127.0.0.1:8000/empty_file", nullptr);
std::cerr << "Exists = " << exists << std::endl;  // prints out 0, which I expects to be 1
```

> I guess this is saying length is never non-0, but it's that actually true?

Hmmm not really, I'm actually trying to handle length = 0 cases

carlopi:
@benfleis, you were talking about FileExits, would you mind making a call on this one? Thanks!

benfleis:
For my money, 0-byte files are files... except when they're not*.

*Even though S3 has no dirs, the directory marker pattern, also respected by e.g. rclone, should potentially return false if the file requested ends with `/`  and empty.

Thus:

- `FileExists("s3://container/empty_file") -> true`
- `FileExists("s3://container/empty_dir/") -> false` IFF size==0-bytes && ends with `/` (and maybe directory marker behavior enabled)

In Azure, I already updated to return true for normal 0 bytes, and implemented the slash check elsewhere (and should add to `FileExists` for consistency).

dentiny:
Thanks @benfleis to chim in!

> Even though S3 has no dirs, the directory marker pattern

Just a supplement for completeness, S3 expression zone supports directory bucket: https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-buckets-overview.html

The "directory" concept is similar to what you've described
> With a hierarchical namespace, the delimiter in the object key is important. The only supported delimiter is a forward slash (/). Directories are determined by delimiter boundaries.

But as I mentioned above
> as long as an OpenFile passes, the file should be decided as "exists", regardless of its file size.

even for "directory" entity in directory buckets, it's `OpenFile`'s responsibility to differentiate directory vs file.

dentiny:
@benfleis Please suggest if there's any change you expect to see in this PR :)

benfleis:
> @benfleis Please suggest if there's any change you expect to see in this PR :)


LGTM, and same intent as what I did in Azure. 

(Any `/` stuff probably should be done simultaneously in a separate effort, across both modules... later.)

benfleis (approved):
👍

PR #213, opened 2026-01-21
Remove `__RESPONSE_STATUS__` from response header

DinosL:
As the title suggests, we should not include `__RESPONSE_STATUS__` in the repsonse header of a request. We keep the content of `__RESPONSE_STATUS__` only for debugging purposes (e.g. failure after a redirect).

### DIFF
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -410,7 +410,8 @@ private:
 		auto status_code = HTTPStatusCode(request_info->response_code);
 		auto response = make_uniq<HTTPResponse>(status_code);
 		if (res != CURLcode::CURLE_OK) {
-			// __RESPONSE_STATUS__ is used for error propagation and debugging. It stores the full HTTP status trail, e.g. in cases of redirection failures, to help trace the sequence of HTTP statuses.
+			// __RESPONSE_STATUS__ is used for error propagation and debugging. It stores the full HTTP status trail,
+			// e.g. in cases of redirection failures, to help trace the sequence of HTTP statuses.
 			if (!request_info->header_collection.empty() &&
 			    request_info->header_collection.back().HasHeader("__RESPONSE_STATUS__")) {
 				response->request_error = request_info->header_collection.back().GetHeaderValue("__RESPONSE_STATUS__");
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -424,6 +424,9 @@ private:
 		response->reason = HTTPUtil::GetStatusMessage(HTTPUtil::ToStatusCode(request_info->response_code));
 		if (!request_info->header_collection.empty()) {
 			for (auto &header : request_info->header_collection.back()) {
+				if (header.first == "__RESPONSE_STATUS__") {
+					continue;
+				}
 				response->headers.Insert(header.first, header.second);
 			}
 		}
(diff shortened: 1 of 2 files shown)


test/sql/copy/csv/test_url_with_plus.test
```diff
@@ -4,6 +4,9 @@
 
 require httpfs
 
+# FIXME: check again later. Something must have changed on the server side and both httplib and curl get code 202 (Accepted).
+mode skip
+
```
carlopi: It seems this test is flaky, but currently is back to working
```suggestion
```

src/httpfs_curl_client.cpp
```diff
@@ -424,6 +424,9 @@ class HTTPFSCurlClient : public HTTPClient {
 		response->reason = HTTPUtil::GetStatusMessage(HTTPUtil::ToStatusCode(request_info->response_code));
 		if (!request_info->header_collection.empty()) {
 			for (auto &header : request_info->header_collection.back()) {
+				if (header.first == "__RESPONSE_STATUS__") {
+					continue;
+				}
```
carlopi, line 431: Could you as well elaborate on:
> // TODO: request error can come from HTTPS Status code toString() value

Ideally changing comment there (and possibly adding one here) on why this behaviour? Thanks!

carlopi (changes requested):
Given that we are here, and we understood this now, could you expand in comment / fixup previous comments?

Thanks

src/httpfs_curl_client.cpp
DinosL: I believe the TODO was left there to make a `toString()` that returns the error message, not just the code.  We can solve this problem by replacing
```
response->request_error = request_info->header_collection.back().GetHeaderValue("__RESPONSE_STATUS__"); 
```  
with
```response->reason = HTTPUtil::GetStatusMessage(HTTPUtil::ToStatusCode(request_info->response_code))```

that I added some lines below for requests that do not fail. This is strictly for more detailed messages for debugging. 


### Some thoughts

This part of the code only gets triggered when there's a **Curl Error** (e.g. `CURLE_COULDNT_RESOLVE_HOST` or `CURLE_TOO_MANY_REDIRECTS`), not for HTTP errors (e.g., `404` or `500`). The idea behind the original code is that for debugging purposes we should keep the HTTP status lines from all redirected requests. 

I did some testing, and even though this can be handy for debugging redirects, it's not strictly necessary since `curl_easy_strerror` already returns the reason for the failure. For example, if we limit the maximum redirects to 1 with `curl_easy_setopt(*curl, CURLOPT_MAXREDIRS, 1L)` and attempt a request that requires multiple redirects, we get the error: `IO Error: Number of redirects hit maximum amount error for HTTP HEAD to '...'`. The curl error message is already informative enough without needing to preserve the intermediate redirect status lines.

carlopi: I think there are 2 paths: either removing `__RESPONSE_STATUS__`-logic completely, or keep it around and update (as per this PR) and also update the comments so that they reflect the new code.
Does this makes sense?

I would think the second option is better, given it offer more informations and more chances of debugging this properly.

DinosL: It makes sense. We keep the `__RESPONSE_STATUS__` and I am adding the `response->reason` to address (and remove) the TODO.

carlopi: I don't think that's needed, I think I would remove the TODO, and reword that as a comment explaining what's the purpose of `__RESPONSE_STATUS__`, to make it clearer to future us.

carlopi (approved):
Thanks!

PR #214, opened 2026-01-22
CURL: init and clenaup appropriately, and avoid static string

carlopi:
Two independent CURL related fixes:
* skip static initialization of `cert_path` (although it's a good idea to cache it)
* restore init & destroy behaviour

### DIFF
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -39,8 +39,6 @@ static std::string SelectCURLCertPath() {
 	return std::string();
 }
 
-static std::string cert_path = SelectCURLCertPath();
-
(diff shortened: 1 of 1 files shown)


carlopi:
@Tmonster: it would be great if you could double check whether this makes any sense to you, if you potentially know of problematic cases.

I think I will add a user side option to disable cleanup at all, just in case.
As per the static initialization of `cert_path`, it's currently unused so should be a NOP, but indeed it would be handy to cache the lookup.

carlopi:
Actually this might be too costly, I have to measure this properly, cleanup should be tied to lifetime of CURLClient.

And also fails Windows CI.

PR #215, opened 2026-01-22
Bump and fix test

carlopi:
Should fix an error on autoloading, unclear why that has been hidden until now.

Actually, I know: httpfs tests where skipped.


DinosL (approved):
Looks good to me :)

PR #216, opened 2026-01-22
Add support for lazily expanding globs to S3FileSystem, and correctly record GET requests made during glob

Mytherin:
This PR implements the lazy glob interface added in https://github.com/duckdb/duckdb/pull/20619 to the S3 globs. Note that the default batch size (1000 keys) is relatively generous - so this is most likely not properly tested by the existing test set. I've run some tests locally with large buckets, and also ran some tests with max key size = 4, and that works - but I'll work on adding some more tests in a follow-up.

This PR also fixes an issue where the `GET` requests made during globs were not correctly recorded due to the `HTTPState` not being initialized correctly.

### DIFF
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -17,7 +17,7 @@ jobs:
     uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@main
     with:
       extension_name: httpfs
-      duckdb_version: 1f9ecad746
+      duckdb_version: 9c2a64a
       ci_tools_version: main
 
   duckdb-stable-deploy:
@@ -27,7 +27,7 @@ jobs:
     secrets: inherit
     with:
       extension_name: httpfs
-      duckdb_version: 1f9ecad746
+      duckdb_version: 9c2a64a
       ci_tools_version: main
       deploy_latest: ${{ startsWith(github.ref, 'refs/heads/v') }}
       deploy_versioned: ${{ startsWith(github.ref, 'refs/heads/v') || github.ref == 'refs/heads/main' }}
--- a/README.md
+++ b/README.md
@@ -1,0 +1,21 @@
+# DuckDB HTTPFS extension
+
+The httpfs extension is an autoloadable extension implementing a file system that allows reading remote/writing remote files. For plain HTTP(S), only file reading is supported. For object storage using the S3 API, the httpfs extension supports reading/writing/globbing files.
+
+## Building & Loading the Extension
+
+The DuckDB submodule must be initialized prior to building.
+
+```bash
+git submodule init
+git pull --recurse-submodules
+```
+
+To build, type
+```
+make
+```
+
+## Testing
+
+Testing uses a local MinIO setup using Docker. See the [testing documentation for more information on how to set this up locally](test).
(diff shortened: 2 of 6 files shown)


carlopi:
Would it make sense to add (behind a configuration flag) something like:
```diff
diff --git a/src/s3fs.cpp b/src/s3fs.cpp
index 65496356453..aca91b0f57b 100644
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1282,6 +1282,7 @@ string AWSListObjectV2::Request(const string &path, HTTPParams &http_params, con
                req_params += "&";
        }
        req_params += "encoding-type=url&list-type=2";
+       req_params += "&max-keys=10";
        req_params += "&prefix=" + S3FileSystem::UrlEncode(parsed_url.key, true);
 
        auto header_map =
```
Then this is tested on different (potentially SQL-provided) values.

I have tested basically:
```sql
CALL enable_logging('HTTP');
.timer on
FROM glob('s3://coiled-datasets/timeseries/20-years/parquet/*.parquet');
```
And with default cap (1000) it gives 1.2 s and 2 requests, while limited to 10 it does 51s and 110 requests.
About as expected.

Only thing I noticed, it seems the Globbing is not very responsive on this setting, unsure if something could be improved there (like returning control more often / or checking interrupt flag).

Mytherin:
I experimented with it a bit and I ended up just adding a test with a bunch of files through MinIO (2000 files) - I'll push it in a follow-up PR. I think that's sufficient as the code is not actually that complex.

The main complexity in the code now is the support for `CommonPrefixes`. I think this can actually be removed. Per [the AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html#API_ListObjects_ResponseSyntax):

> [CommonPrefixes](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html#API_ListObjects_ResponseSyntax)
> All of the keys (up to 1,000) rolled up in a common prefix count as a single return when calculating the number of returns.
>
> A response can contain CommonPrefixes only if you specify a delimiter.
>
> CommonPrefixes contains all (if there are any) keys between Prefix and the next occurrence of the string specified by the delimiter.
>
> CommonPrefixes lists keys that act like subdirectories in the directory specified by Prefix.
>
> For example, if the prefix is notes/ and the delimiter is a slash (/), as in notes/summer/july, the common prefix is notes/summer/. All of the keys that roll up into a common prefix count as a single return when calculating the number of returns.

We used to have `delimiter` support in the glob, but that seemed to be unused and hence I ended up removing it in this PR. I think we can also remove support for CommonPrefixes then as we no longer support passing a delimiter to this API, but maybe I'll just leave it in for now. It's not super clear to me what the purpose of this was originally and/or if this will ever be needed again.

carlopi:
I think the `delimiter` would be better to keep, but then there is some extra logic that needs fixing.

As a testcase (2 globs)
```sql
FROM glob('s3://coiled-datasets/timeseries/*-years/parquet/*.parquet');
```
then without delimiter, it seems like it's listing all files matching `s3://coiled-datasets/timeseries/*` (around 46 requests -> 46000 files), and filtering them afterwards.

While with delimiter, it's looking like it's listing directories recursing one globbing level after the other, and only then recursing at the next level.
There is something off since it's still doing extra requests that are not needed (like `GET` to list on `https://coiled-datasets.s3.amazonaws.com/?encoding-type=url&list-type=2&delimiter=%2F&prefix=timeseries%2Fspark%2F`), but it seems to me more sensible to do recursive globbing.


Basically, currently:
```
FROM glob('s3://coiled-datasets/timeseries/*-years/parquet/*.parquet');
--- and
FROM glob('s3://coiled-datasets/timeseries/*');
```
are executed with the same logic, while the first has more chances to prune search.

Mytherin:
I think that's related to the pushdown into globs which we don't (yet) do. Maybe something to revisit in the future indeed

PR #218, opened 2026-01-22
Add back handling for use_delimiter, making recursive globs faster

carlopi:
Test is:
```sql
CALL enable_logging('HTTP');
.timer on
FROM glob('s3://coiled-datasets/timeseries/*-years/parquet/*.parquet');
FROM duckdb_logs_parsed('HTTP') SELECT count(*);
```
that takes now about 2s and 4 requests (vs 42s and 46 requests, both main and v1.4-andium behaviour).

One can look at the difference via `FROM duckdb_logs_parsed('HTTP') SELECT request.url;` while toggling `SET allow_s3_recursive_globbing = false;`/`SET allow_s3_recursive_globbing = true;`:
```
│ /?encoding-type=url&list-type=2&prefix=timeseries%2F                                                                                                                                                                                         │
│ /?continuation-token=11e5rOWab38dy2uXgMSmx%2B8xAYGzeScEsqHifLHfxqNEbKLf7NiBeAbL38fFcRlk9sLuZlLf98zbkf2u0fYKO3HhyhwqXCgxU&encoding-type=url&list-type=2&prefix=timeseries%2F                                                                  │
│ /?continuation-token=1ag6ez2QkIXhAZgecXjczrkErG75B8loUGMlTuIhvt3Jt4FoBOKK1fKy3XDCA8TdqdSeRxv%2B7kMBUaeBfgEhdaw%3D%3D&encoding-type=url&list-type=2&prefix=timeseries%2F                                                                      │
│ /?continuation-token=1HSOEN%2FRH2Zc5pXJh6Jcx7GWiV9erWtkJrWSQmhnHceYuCDeVkDouFT86aH6vsrFQhORFDuwWaQxQVzhdDAY7dw%3D%3D&encoding-type=url&list-type=2&prefix=timeseries%2F                                                                      │
│ /?continuation-token=1SXO9x37RvK0%2FhCJE%2BbxPoBikpdmjd5%2FSXvCbq%2BZR2ESMnrk1B78qPkMoACmt%2BNtSP2Z5YYhnPXX8acYsVy7NMQ%3D%3D&encoding-type=url&list-type=2&prefix=timeseries%2F                                                              │
│ /?continuation-token=1ZkHF3sIKDud%2BIIEPsRksgRB2jpF6UMvDqpUf%2FgDL%2BsyoQfoBiG2kwgK0Y2IO0rYry4n%2FhtabJcxNrNVSeb8ATA%3D%3D&encoding-type=url&list-type=2&prefix=timeseries%2F                                                                │
│ /?continuation-token=1mj9qKmmG3HiFLDNSX9PEHrqj8fLr4SDMK55W%2Bvg%2BPyq24UMMZiVOR5Da8l3dmNDNosfPu2clm3M8n8LDu32OHg%3D%3D&encoding-type=url&list-type=2&prefix=timeseries%2F                                                                    │
│ /?continuation-token=1swIROhM2IyFW9SKfJLmF0xiT%2BhhguFNSj4O6wRSUbeie5NtMoCXhBVNGQXh6g4AeXo4hOv1wYf4PvtR2jdWiYQ%3D%3D&encoding-type=url&list-type=2&prefix=timeseries%2F                                                                      │
│ /?continuation-token=1qZxLDva9BngXXDKnFdjKyn06%2Fyhzv%2FqUkluNOVZl6n8n8CjwLXAZn8qwzAz6vQ%2FsE0vbmb%2Fckb4LfgOR6VVBjg%3D%3D&encoding-type=url&list-type=2&prefix=timeseries%2F                                                                │
│ /?continuation-token=13%2F6p41lFEClH%2FJkk7I2P5HZb2Oj7%2B9E0tLkSjixyFaEEemjiWHt%2FGow5hKpxYN0v4SjUoOCc9ZVqBsIS6BYeOQ%3D%3D&encoding-type=url&list-type=2&prefix=timeseries%2F                                                                │
.... more like this, all continuation happens at the level of the first glob
```
vs
```
│ /?delimiter=%2F&encoding-type=url&list-type=2&prefix=timeseries%2F                                                                                                                                             │
│ /?delimiter=%2F&encoding-type=url&list-type=2&prefix=timeseries%2F20-years%2F                                                                                                                                  │
│ /?delimiter=%2F&encoding-type=url&list-type=2&prefix=timeseries%2F20-years%2Fparquet%2F                                                                                                                        │
│ /?continuation-token=1EsJlUTGcNUrzxV5MykA9qEVBV03ndN0WG1uCUGhwxm2f1FG6cvnPOQdTltlhGgTalu5xggNk%2FJwl2ECbKppmtaYGHnrpgfHo&delimiter=%2F&encoding-type=url&list-type=2&prefix=timeseries%2F20-years%2Fparquet%2F │
```
where one can see the directories are scanned recursively (first `timeseries%2F` obtaining the next level of directories, then one level deeper at `timeseries%2F20-years%2F`, then at timeseries%2F20-years%2Fparquet%2F with a continuation call since there were more than a 1000 there).

This changes from basically listing anything with a common root, to listing by directory and recursively expanding (and matching accordingly).

This will worsen performances in structures where just listing it's faster then directory traversing, say a structure like:
```
1
└── 2
    └── 3
        └── 4
            └── 5
                └── 6
                    ├── file
                    ├── file2
                    └── file3

```
then 6 hops are required before finally listing at depth 6 (vs a single request given there are less than 1000 files).

This will worsen performance by a number of requests that's equal to the total number of nodes in the trees.

Worse case for current globbing methods are partitioned files, where there are many files per each directory, so the extra cost of walking the directories is limited, while the number of request will growth with the total number of files in the folder.

Approximate complexities are:
* recursive globbing: `O(number of internal directories on the path) + O(result set)`
* regular globbing: `O(total number of files matching the prefix)`

A more extreme example, in the first case, it's a query like:
```
FROM glob('s3://coiled-datasets/timeseries/*-years/i-do-not-exist/*.parquet');
```
where if `i-do-not-exist` is not there, 2 network trip are enough, vs the 46 fixed cost that it's still needed in any case in the other configuration.

I think having number of roundtrip bounded by result set it's better over the board, so I would propose that as default.

### DIFF
--- a/src/httpfs_extension.cpp
+++ b/src/httpfs_extension.cpp
@@ -82,7 +82,7 @@ static void LoadInternal(ExtensionLoader &loader) {
 	                          LogicalType::BOOLEAN, Value(false));
 	config.AddExtensionOption("s3_requester_pays", "S3 use requester pays mode", LogicalType::BOOLEAN, Value(false));
 	config.AddExtensionOption(
-	    "allow_s3_recursive_globbing",
+	    "s3_allow_recursive_globbing",
 	    "Whether globs on S3-like storage are optimized with recursive strategy (alterative is listing)",
 	    LogicalType::BOOLEAN, Value(true));
 
--- a/src/httpfs_extension.cpp
+++ b/src/httpfs_extension.cpp
@@ -81,6 +81,10 @@ static void LoadInternal(ExtensionLoader &loader) {
 	config.AddExtensionOption("s3_url_compatibility_mode", "Disable Globs and Query Parameters on S3 URLs",
 	                          LogicalType::BOOLEAN, Value(false));
 	config.AddExtensionOption("s3_requester_pays", "S3 use requester pays mode", LogicalType::BOOLEAN, Value(false));
+	config.AddExtensionOption(
+	    "allow_s3_recursive_globbing",
+	    "Whether globs on S3-like storage are optimized with recursive strategy (alterative is listing)",
+	    LogicalType::BOOLEAN, Value(true));
 
 	// S3 Uploader config
 	config.AddExtensionOption("s3_uploader_max_filesize", "S3 Uploader max filesize (between 50GB and 5TB)",
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1333,7 +1333,14 @@ bool S3GlobResult::ExpandNextPath() const {
 		if (!common_prefixes.empty()) {
 			throw InternalException("We have common prefixes but we are doing a top-level request");
 		}
-		const bool use_recursive_glob = !StringUtil::Contains(parsed_s3_url.key, "**");
+
+		Value value;
+		bool allow_s3_recursive_globbing;
+		if (FileOpener::TryGetCurrentSetting(opener, "allow_s3_recursive_globbing", value)) {
+			allow_s3_recursive_globbing = value.GetValue<bool>();
+		}
+
+		const bool use_recursive_glob = !StringUtil::Contains(parsed_s3_url.key, "**") && allow_s3_recursive_globbing;
 		// issue the main request
 		string response_str = AWSListObjectV2::Request(shared_path, *http_params, s3_auth_params,
 		                                               main_continuation_token, use_recursive_glob);
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1201,23 +1201,24 @@ void S3FileSystem::Write(FileHandle &handle, void *buffer, int64_t nr_bytes, idx
 static bool Match(vector<string>::const_iterator key, vector<string>::const_iterator key_end,
                   vector<string>::const_iterator pattern, vector<string>::const_iterator pattern_end, bool completed) {
 
-		if (key == key_end && !completed) {
-			return true;
-		}
+	if (key == key_end && !completed) {
+		return true;
+	}
 
 	while (key != key_end && pattern != pattern_end) {
 		if (*pattern == "**") {
 			if (std::next(pattern) == pattern_end) {
 				return true;
 			}
-			pattern ++;
+			pattern++;
 			while (key != key_end) {
 				if (Match(key, key_end, pattern, pattern_end, completed)) {
 					return true;
 				}
 				key++;
 			}
-			if (!completed) return true;
+			if (!completed)
+				return true;
 			return false;
 		}
 		if (!Glob(key->data(), key->length(), pattern->data(), pattern->length())) {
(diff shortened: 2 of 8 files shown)


Mytherin:
Thanks - bringing back the delimiter is fine by me (although I don't fully understand what this is doing that is causing this speed-up).

One thing - when I run this locally I get an error:

```sql
set s3_region='us-east-2';
FROM glob('s3://coiled-datasets/timeseries/*-years/parquet/*.parquet');
```

```
HTTP Error:
HTTP GET error reading 's3://coiled-datasets/timeserieshttps://coiled-datasets.s3.us-east-2.amazonaws.com/?encoding-type=url&list-type=2&prefix=timeseries%2F&delimiter=%2F' in region 'us-east-2' (HTTP 403 Forbidden)

Authentication Failure - this is usually caused by invalid or missing credentials.
* Credentials are provided, but they did not work.
* See https://duckdb.org/docs/stable/extensions/httpfs/s3api.html

LINE 1: FROM glob('s3://coiled-datasets/timeseries/*-years/parquet/*...
             ^
```

There seems to be something going wrong with concatenation of URLs (`s3://coiled-datasets/timeserieshttps://...`). Are you seeing this locally as well?

carlopi:
I don't reproduce, this smells like you might have credentials or region enabled? That it's a longstanding (and weird) issue.

Can you check `SELECT current_setting('s3_access_key_id');` (and family)?

carlopi:
As to why this is faster, basically in the case of multiple folders, like:
```
my-dir/file1.parquet
my-dir/file2.parquet
other-dir/a/file.parquet
other-dir/a/file2.parquet
zz-dir/z/file2.parquet
other-folder/a/file2.parquet
```
Say that you glob on `*-dir/a/*.parquet`, before the PR it would list all files in the bucket, and then select 3rd and 4th.
After it would:
1. list everything up to the delimiter (so getting back `dir/` + `other-folder/`)
2. iterating on `my-dir`, that matches, now listing on `my-dir/a/*.parquet` will get 0 files.
3. iterationg on `other-dir`, that matches, now it should list every file in `dir/a/*.parquet`, getting 2 files.
4. iterating on `zz-dir`, that matches
5. iterating on `other-folder`, that do not matches

Now that I write it down, it can be more expensive since every level gets an extra listing, advantage it should be that those are parallelizable, while the single full folder scan it's not parallelizable.

src/s3fs.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include "create_secret_functions.hpp"
 
+#include <iostream>
```
carlopi, line 24:
```suggestion
```

carlopi:
@lnkuiper & @Mytherin: I think this is ready to go in, and would be great if either of you could make a pass, given comments or approval to the PR.

Short version is that I would like recursive globbing, where folders are expanded one-by-one as done in native, the default choice.

Overhead it's number of visited folders, vs worse case of current behavior where the number of requests scale with number of files matching initial glob (vs result set).
I think that while this will perform worse in test setup, it will give consistent behaviour across larger [hive-]partitioned production scenarios.

Main points that are not tight (and I am aware of that):

### How to decide strategy
Logic is as follow:
* if option it's disabled, via `SET allow_s3_recursive_globbing = false`, then use always current listing-style globbing
* if there is a "**" anywhere in the glob pattern, use listing-style globbing (that it's the current default)
* otherwise, use recursive-style globbing

There are slightly more cases where one can prove the old globbing is better, and employing a mix in certain cases (like `s3://bucket/*/subfolder/**` would benefit from doing first part recursive and second via listing), but in the sake of simplicity and robustness this seems the better way forward.

### Which one should be the default
I think recursive style is the correct default, but open to discussion.

src/httpfs_extension.cpp
```diff
@@ -81,6 +81,10 @@ static void LoadInternal(ExtensionLoader &loader) {
 	config.AddExtensionOption("s3_url_compatibility_mode", "Disable Globs and Query Parameters on S3 URLs",
 	                          LogicalType::BOOLEAN, Value(false));
 	config.AddExtensionOption("s3_requester_pays", "S3 use requester pays mode", LogicalType::BOOLEAN, Value(false));
+	config.AddExtensionOption(
+	    "allow_s3_recursive_globbing",
```
lnkuiper: Can we rename this to `s3_allow_recursive_globbing` so it's prefixed with `s3_` like the other options?

lnkuiper (changes requested):
I have left one comment below, and my opinion is that this is most likely the more sane default behaviour, so I'm in favour of enabling it by default.

src/s3fs.cpp
```diff
@@ -1311,9 +1333,17 @@ bool S3GlobResult::ExpandNextPath() const {
 		if (!common_prefixes.empty()) {
 			throw InternalException("We have common prefixes but we are doing a top-level request");
 		}
+
+		Value value;
+		bool allow_s3_recursive_globbing = true;
+		if (FileOpener::TryGetCurrentSetting(opener, "allow_s3_recursive_globbing", value)) {
+			allow_s3_recursive_globbing = value.GetValue<bool>();
```
lnkuiper, line 1340: Also, one more nit: I think there is a more ideomatic way to get extension settings now that Mark has refactored them. Example: https://github.com/duckdb/duckdb/pull/20519/changes#diff-55b7a234022192f0d08c7e50ec9b7cf7b1138ea217cfbb2b1c7c58913040d6d5R36

carlopi: I am not sure I see it.

carlopi:
Thanks for reviews, should be good to go in.

PR #219, opened 2026-01-22
Revert listobjectv2 fix and consequently CURL as default

carlopi:


Mytherin:
The CI errors I have also been hitting locally - they are caused by tests not being re-entrant (e.g. `test/sql/copy/s3/starstar.test` checks that a directory is empty, and then afterwards writes to that directory, so running the test twice makes the second run fail).

carlopi:
Yes, that I will fix by doing explicitly cleanup, it's indeed annoying, and in a previous version I mistakenly swapped CI around.

carlopi:
Outdated

PR #223, opened 2026-01-26
CURL: Refactor the client to better use the curl API for constructing URLs

DinosL:
This PR refactors the `curl` client to make url handling cleaner and more consistent to `httplib` client implementation.

First, it addresses a TODO by actually using `proto_host_port`. This value is now used to construct a base URL via libcurl’s API, which fixes issues caused by incomplete or incorrectly assembled URLs (e.g. https://github.com/duckdblabs/duckdb-internal/issues/6041)

Second, the code is now more “curl-ified.” URL construction and manipulation are delegated to the library instead of relying on our own functions. As part of this, the custom `EncodeSpaces` helper is removed. When setting URLs via `curl_url_set()` (using `CURLUPART_URL` or `CURLUPART_PATH`), the library correctly handles URL encoding internally, making the manual encoding unnecessary. 

Also, fixes issues in error messages (e.g. https://github.com/duckdblabs/duckdb-internal/issues/7211)

### DIFF
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -115,7 +115,8 @@ static idx_t httpfs_client_count = 0;
 class HTTPFSCurlClient : public HTTPClient {
 public:
 	HTTPFSCurlClient(HTTPFSParams &http_params, const string &proto_host_port) {
-		// FIXME: proto_host_port is not used
+		base_url = curl_url();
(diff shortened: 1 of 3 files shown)


carlopi:
Thanks a ton, this makes a lot of sense to me!

carlopi (approved):

PR #227, opened 2026-01-27
CURL: Use CA cert file when available

DinosL:
Basically what the title says.

Fixes: https://github.com/duckdblabs/duckdb-internal/issues/7232


carlopi (approved):
Thanks!

PR #232, opened 2026-01-28
CURL: avoid re-initializing the CURL client completely

carlopi:
Re-initialize CURL object only in limited cases:
* if null
* or if provided certificate path has changed
* or if provided bearer_token has changed

This basically means that repeated calls of Initialize(), potentially intertwined with requests, should now be "cheap", and enabling further work in caching HTTPClients

### DIFF
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -39,8 +39,6 @@ static std::string SelectCURLCertPath() {
 	return std::string();
 }
 
-static std::string cert_path = SelectCURLCertPath();
-
 static size_t RequestWriteCallback(void *contents, size_t size, size_t nmemb, void *userp) {
 	size_t totalSize = size * nmemb;
 	std::string *str = static_cast<std::string *>(userp);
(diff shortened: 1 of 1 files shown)


src/httpfs_curl_client.cpp
```diff
@@ -460,6 +473,8 @@ class HTTPFSCurlClient : public HTTPClient {
 	optional_ptr<HTTPState> state;
 	unique_ptr<RequestInfo> request_info;
 	CURLU *base_url = nullptr;
+	string stored_bearer_token;
```
DinosL, line 476: `stored_bearer_token` is initialized to an empty string and its value is never changed. If I am not missing something, why do we need it?

carlopi: I think this was correct in a past version, could you take another look in the new version I had force-pushed? Thanks!

carlopi: Should be handled via https://github.com/duckdb/duckdb-httpfs/pull/235, thanks for checking again

PR #238, opened 2026-01-30
Skip test on Windows, to be investigated `curl` global options

carlopi:
Working theory it's that curl is now used both in duckdb-httpfs AND duckdb-spatial (via GDAL) and there is some shared global state. To be investigated, but for now skipping Windows.

Note that the problem is very borderline: if your machine has (via configuration) already a certificate verification method, that can't be overridden. This is relevant to look into, but the happy path where one do not mess with certificate will keep working, what's off (in some Windows configuration only) is the capability of overriding certificates.


carlopi:
Test-only, CI is irrelevant here.

PR #239, opened 2026-02-02
HTTPClient caching across FileHandles

carlopi:
Currently `HTTPClientCache`s are per FileHandle, after this PR they each FileHandle has a `shared_ptr<HTTPClientCache>` that allows to share the clients across the same base-url.

An LruCache is kept to reachable `HTTPClientCache`'s, that allows repeated network call, even across `FileHandle`'s, to avoid unnecessary network handshakes.

This builds on top of LruCache made available in `duckdb/duckdb` in https://github.com/duckdb/duckdb/pull/20157 (and generalized in https://github.com/duckdb/duckdb/pull/20757 and the infrastructure to re-initialized clients, recently touched on via https://github.com/duckdb/duckdb-httpfs/pull/232.

Ownership model is fully RAII, thanks to `shared_ptr`'s, note that this means that there might be still alive `HTTPClientCache`, say a very long lived `FileHandle`, but those are not reachable anymore.
This can be improved on, but for now PR to improve current status and add infrastructure.
Cache size is fixed at 256, arguably it should be configurable.

This should somewhat visibly improve performances of repeated operations to same `BaseUrl()`, but apart from timing this should not bring observable differences.
CI-wise, main goal here is checking whether it's a pass.

### DIFF
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -24,6 +24,15 @@
 
 namespace duckdb {
 
+	ClientOptions::ClientOptions(HTTPFileHandle &handle) {
+		identifier = "";
+		identifier += handle.params->http_proxy + " ";
+		identifier += handle.params->http_proxy_username + " ";
+		identifier += handle.params->http_proxy_password + " ";
+		//identifier += handle.params->http_proxy_port + " ";
+		identifier += handle.params->http_util.GetName();
+	}
+
 shared_ptr<HTTPUtil> HTTPFSUtil::GetHTTPUtil(optional_ptr<FileOpener> opener) {
 	if (opener) {
 		return opener->GetHTTPUtil();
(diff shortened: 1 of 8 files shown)


src/httpfs.cpp
```diff
@@ -377,6 +377,29 @@ HTTPFileHandle::HTTPFileHandle(FileSystem &fs, const OpenFileInfo &file, FileOpe
 		}
 	}
 }
+
+shared_ptr<HTTPClientCache> HTTPFileSystem::GetOrCreateClientCache(const string &path) {
+	lock_guard<mutex> lock(client_cache_map_lock);
+
+	if (auto existing = lru_client_cache.Get(path)) {
```
lnkuiper: I get what you're doing here, and it looks correct, but we try to avoid assignment in `if` conditions because it's error-prone. Could you split this up?

test/sql/json/table/internal_issue_6807.test_slow
```diff
@@ -15,4 +15,4 @@ CREATE TABLE T AS FROM 'https://data.gharchive.org/2023-02-08-0.json.gz';
 query I
 SELECT count(*) FROM duckdb_logs_parsed('HTTP') WHERE request.type = 'GET' GROUP BY request.type;
 ----
-9
+8
```
lnkuiper: I think this change is already in a patch in the duckdb repo on `v1.5-variegata`. Not a problem if you add the change here too as long as CI is happy

lnkuiper (commented):
Looks good! I have left two minor comments below, and I have a question about testing. Our current HTTP stats (and logs) currently don't show # of handshakes:
```
┌─────────────────────────────────────┐
│┌───────────────────────────────────┐│
││         HTTPFS HTTP Stats         ││
││                                   ││
││           in: 131.2 MiB           ││
││            out: 0 bytes           ││
││              #HEAD: 1             ││
││              #GET: 9              ││
││              #PUT: 0              ││
││              #POST: 0             ││
││             #DELETE: 0            ││
│└───────────────────────────────────┘│
└─────────────────────────────────────┘
```
Would it be possible to measure this and add a test so we can guarantee that this continues to work?

carlopi:
Thanks, handled the comments, on adding to "HTTPFS HTTP Stats" I am a bit conflicted, since this should be mostly invisible to users, basically it's an optimisation that should just happen, and all other signal are mostly noisy.

I could think of adding a log type, that could be handy, but even then unsure if that's required in this PR.

I tested with a proxy, that logs whether actual handshake happened or not, and results are as expected, that is first round of connections (about 1 per thread) perform handshake, while next connections will re-use pre-initialized clients.

lnkuiper:
Superseded by https://github.com/duckdb/duckdb/pull/25358

PR #243, opened 2026-02-09
add S3 version id pinning to httpfs

OneCyrus:
the current implementation in HTTPFS tries to always access the latest version of objects in S3. this might result in downloading parts of a file from different versions. This PRs fixes this by using the version id if versioning is enabled on the S3 bucket. It does this by storing the version id after the first request to the object. Subsequent file access uses this version id if available and can download the remaining parts even when the file changed mid download on S3.

basically it captures the `x-amz-version-id` http header:
```
HTTP/1.1 200 OK
x-amz-version-id: 3HL4kqCxf3vjVBH40Nrjfkd
ETag: "abc123..."
Content-Length: 987654321
```

and adds it as a query parameter to requests

```
GET /my-object.parquet?versionId=3HL4kqCxf3vjVBH40Nrjfkd HTTP/1.1
Host: my-bucket.s3.amazonaws.com
Range: bytes=0-1048575
```

I put this feature behind a feature flag which disabled by default. To try this enable the flag 
`SET s3_version_id_pinning = true;`

this fixes the issue https://github.com/duckdb/duckdb/issues/20533

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -802,9 +802,16 @@ unique_ptr<HTTPResponse> S3FileSystem::HeadRequest(FileHandle &handle, string s3
 }
 
 unique_ptr<HTTPResponse> S3FileSystem::GetRequest(FileHandle &handle, string s3_url, HTTPHeaders header_map) {
-	auto auth_params = handle.Cast<S3FileHandle>().auth_params;
+	auto &s3_handle = handle.Cast<S3FileHandle>();
+	auto auth_params = s3_handle.auth_params;
 	auto parsed_s3_url = S3UrlParse(s3_url, auth_params);
-	string http_url = parsed_s3_url.GetHTTPUrl(auth_params);
+
+	string query_string;
+	if (!s3_handle.version_id.empty()) {
+		query_string = "versionId=" + UrlEncode(s3_handle.version_id, true);
+	}
+
+	string http_url = parsed_s3_url.GetHTTPUrl(auth_params, query_string);
 
 	HTTPHeaders headers;
 	if (IsGCSRequest(s3_url) && !auth_params.oauth2_bearer_token.empty()) {
@@ -813,7 +820,7 @@ unique_ptr<HTTPResponse> S3FileSystem::GetRequest(FileHandle &handle, string s3_
 		headers["Host"] = parsed_s3_url.host;
 	} else {
 		// Use existing S3 authentication
-		headers = CreateS3Header(parsed_s3_url.path, "", parsed_s3_url.host, "s3", "GET", auth_params, "", "", "", "");
+		headers = CreateS3Header(parsed_s3_url.path, query_string, parsed_s3_url.host, "s3", "GET", auth_params, "", "", "", "");
 	}
 
 	return HTTPFileSystem::GetRequest(handle, http_url, headers);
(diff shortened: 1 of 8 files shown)


carlopi:
Couple of questions:
* has this been tested with actual AWS S3 bucket? I will also do so as part of review process, but it can be handy since `sigv4` code it's different between providers, and AWS can be picky about ordering of parameters.
* If I understand correctly it means that one will be tied to the older file, what's the way to opt-in to checking whether the file has been updated? (I guess removing the cache?) Could a test be added where a file is written, then partially read (like metadata only), then overwritten, then read again, and showing that the results are from the first write?

Something like:
```sql
CREATE TABLE T1 as SELECT random() FROM range(123124);
CREATE TABLE T2 as SELECT random() FROM range(123124);
COPY T1 TO 's3://bucket/file.parquet';
SET s3_version_id_pinning = true;
SELECT glob('s3://bucket/file.parquet');
COPY T2 TO 's3://bucket/file.parquet' (OVERWRITE);
FROM 's3://bucket/file.parquet';   --- this should be T1
---- clean caches, how?
FROM 's3://bucket/file.parquet';   --- this should be T2
```

OneCyrus:
> Couple of questions:
> 
> * has this been tested with actual AWS S3 bucket? I will also do so as part of review process, but it can be handy since `sigv4` code it's different between providers, and AWS can be picky about ordering of parameters.

i couldn't verify it with a real S3 bucket yet as we only have MINIO in use.

> * If I understand correctly it means that one will be tied to the older file, what's the way to opt-in to checking whether the file has been updated? (I guess removing the cache?) Could a test be added where a file is written, then partially read (like metadata only), then overwritten, then read again, and showing that the results are from the first write?

yes, that's my current understanding. the metadata cache in the released version leads to the etag issues mentioned in the referenced issue as it has metadata which don't work with the content the server returns. the version id pinning here is bound to the same lifecycle as the metadata cache. i guess the proper solution for metadata cache is some kind of cache invalidation (looks like the current cache implementation has no retention policy).

I just tried to add tests but hit issues as the minio version used in this repo is quite old (from 2021) and has some issues with versioning and etags. tried to update to a newer minio version and my tests run but hitting issues with other tests. any chance you could help to update minio in the tests to a newer version?

OneCyrus:
@carlopi i added the tests but as mentioned above the current old minio version used in the test environment doesn't work correctly

carlopi:
I bumped the version in https://github.com/duckdb/duckdb-httpfs/pull/247, let's see if that just works, if so you might consider merging / rebasing / adding that commit to this PR.

carlopi:
@OneCyrus: I took the liberty to rebasing on `main`, where I think it's more appropriate to introduce such a change, and it will also be tested properly via `Minio` setup. PR is at https://github.com/duckdb/duckdb-httpfs/pull/248

src/s3fs.cpp
```diff
@@ -802,9 +802,16 @@ unique_ptr<HTTPResponse> S3FileSystem::HeadRequest(FileHandle &handle, string s3
...
 
 unique_ptr<HTTPResponse> S3FileSystem::GetRequest(FileHandle &handle, string s3_url, HTTPHeaders header_map) {
-	auto auth_params = handle.Cast<S3FileHandle>().auth_params;
+	auto &s3_handle = handle.Cast<S3FileHandle>();
+	auto auth_params = s3_handle.auth_params;
 	auto parsed_s3_url = S3UrlParse(s3_url, auth_params);
-	string http_url = parsed_s3_url.GetHTTPUrl(auth_params);
+
+	string query_string;
+	if (!s3_handle.version_id.empty()) {
```
Tmonster, line 810: should we also check `s3_version_id_pinning` here? Potentially a request was made with it on at first, then it was disabled, but the cache still has a valid version_id?

Tmonster: Should we also check that the s3_url is not a GCS request with `IsGCSRequest()`?

Tmonster: Better yet, if pinning is enabled and the request goes to a GCS/R2 bucket, maybe we should throw an error or warning so users don't assume pinning is valid for GCS/R2 reads?

Tmonster (commented):
Should we also be version pinning HEAD requests? 
I imagine it is possible that a head request is made for some V1 version? Then the get request records the V2 version id? If that happens I can imagine the length information returned by the HEAD request might not match information from the GET request?

Also, it seems like a user needs to hold this just write in order to read back data they just wrote. I.e a situation like the following requires careful setting of the new settings, Do we want to maybe clear the version pinning on `PUT` or `POST` to a file that is in the version pinning cache?
```
COPY T1 TO 's3://bucket/file.parquet';
SELECT * FROM 's3://bucket/file.parquet';
```

src/s3fs.cpp
OneCyrus: my understanding is that GCS and R2 are using the S3 API as well? So it should work the same as any other S3 storage.

OneCyrus: R2 even strongly recommends to do it

image

OneCyrus:
feature introduced in PR #248

src/s3fs.cpp
carlopi: @Tmonster: thanks for the review, I think concern is valid but given this it's a sort of standard, and that we will enable only for server that will send back `x-amd-version-id` headers, I think we are good.

Basically, if a server send that header, I think it's safe to assume they want to opt-in.
We could expand this to other dialect, other provider might use different naming (for example, GCS uses the `generation: <generation_number>` header I think), but I think this should behave as expected.

PR #247, opened 2026-02-10
Bump minio image from 2021-11-03 to 2025-08-13

carlopi:
Bump to latest available `minio` and `mc` docker images, minor syntactic changes and likely relevant improvements.

PR #248, opened 2026-02-10
OneCyrus's add S3 version id pinning to httpfs [on main]

carlopi:
From the original PR, at https://github.com/duckdb/duckdb-httpfs/pull/243, with minor fixups to rebased to `main` that includes a bump in minio to latest available image (https://github.com/duckdb/duckdb-httpfs/pull/247)

-----

the current implementation in HTTPFS tries to always access the latest version of objects in S3. this might result in downloading parts of a file from different versions. This PRs fixes this by using the version id if versioning is enabled on the S3 bucket. It does this by storing the version id after the first request to the object. Subsequent file access uses this version id if available and can download the remaining parts even when the file changed mid download on S3.

basically it captures the `x-amz-version-id` http header:
```
HTTP/1.1 200 OK
x-amz-version-id: 3HL4kqCxf3vjVBH40Nrjfkd
ETag: "abc123..."
Content-Length: 987654321
```

and adds it as a query parameter to requests

```
GET /my-object.parquet?versionId=3HL4kqCxf3vjVBH40Nrjfkd HTTP/1.1
Host: my-bucket.s3.amazonaws.com
Range: bytes=0-1048575
```

I put this feature behind a feature flag which disabled by default. To try this enable the flag 
`SET s3_version_id_pinning = true;`

this fixes the issue https://github.com/duckdb/duckdb/issues/20533

### DIFF
--- a/test/sql/copy/s3/version_id_pinning.test
+++ b/test/sql/copy/s3/version_id_pinning.test
@@ -1,0 +1,91 @@
+# name: test/sql/copy/s3/version_id_pinning.test
+# description: Test S3 version ID pinning: reads remain pinned to the original object version even after the file is overwritten.
+# group: [s3]
+
+require parquet
+
+require httpfs
+
+require-env S3_TEST_SERVER_AVAILABLE 1
+
+require-env AWS_DEFAULT_REGION
+
+require-env AWS_ACCESS_KEY_ID
+
+require-env AWS_SECRET_ACCESS_KEY
+
+require-env DUCKDB_S3_ENDPOINT
+
+require-env DUCKDB_S3_USE_SSL
+
+# override the default behaviour of skipping HTTP errors and connection failures: this test fails on connection issues
+set ignore_error_messages
+
+# disable external file cache to avoid interference from locally cached file content
+statement ok
+SET enable_external_file_cache=false;
+
+# Create two tables with different row counts so we can distinguish them
+statement ok
+CREATE TABLE T1 AS SELECT i FROM range(100) tbl(i);
+
+statement ok
+CREATE TABLE T2 AS SELECT i FROM range(200) tbl(i);
+
+# Write T1 to S3 (bucket must have versioning enabled)
+statement ok
+COPY T1 TO 's3://test-bucket-public/root-dir/version_pinning/test.parquet';
+
+# Enable version pinning and the global metadata cache.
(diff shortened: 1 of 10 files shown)


carlopi:
Currently default is, as per original PR, to `false`, but one open question I have, either short or medium time, we should improve and turn this on by default.

All cases where duckdb does reads and writes they should work the same, and there should be more cases where we will eb robust to out-of-band writes.

Note this feature needs to be opt-in by whoever manages a given S3 bucket, and has no cost or complexity if the versioning it's not enabled.

I have a branch switching default running CI at https://github.com/carlopi/duckdb_httpfs/actions/runs/21872198036 and https://github.com/carlopi/duckdb_httpfs/actions/runs/21872198312

carlopi:
@OneCyrus: any feedback from your side? On the rebased PR or on what the default should be?

OneCyrus:
@carlopi that looks good. as for default i think it should be true as well. just didn't want to introduce a breaking change so I put it behind a flag.

test/sql/copy/s3/version_id_pinning.test
```diff
@@ -0,0 +1,91 @@
...
+# the per-context cache, not the global cache.
+statement ok
+COPY T2 TO 's3://test-bucket-public/root-dir/version_pinning/test.parquet';
+
+# Re-enable the global metadata cache. It still holds the V1 entry.
+statement ok
+SET enable_http_metadata_cache=true;
+
+# Read the file again. The global cache provides the V1 version_id,
+# so the GET request includes ?versionId=V1 and returns the original T1 data.
```
samansmink, line 80: Can we maybe use the logger (`call enable_logging('http')`) to inspect the actual requests made here?

samansmink (commented):
I think this looks good, i have one small comment related to testing

test/sql/copy/s3/version_id_pinning.test
carlopi: Good catch on the testing, since that also revealed that the test was not behaving as expected (at least on Minio test setup non-credentialled access will not get the version) and was basically not testing this feature.

carlopi:
Thanks @OneCyrus, looks to be working great, now enabled by default.

OneCyrus:
@carlopi thank you too for helping me on this PR! and great to see this merged 😎

PR #252, opened 2026-02-12
Reset all relevant fields post usage, and set FOLLOWLOCATION depending on parameters

carlopi:


carlopi:
@Tmonster provided feedback out of band, working also with this change-set in Iceberg testing

PR #258, opened 2026-02-23
Used cached handle with s3 region

jeewonhh:
Use cached entry so that corrected region parameter takes effect in subsequent request. 

Fixes https://github.com/duckdblabs/motherduck/issues/456

### DIFF
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -933,7 +933,7 @@ void HTTPFileHandle::Initialize(optional_ptr<FileOpener> opener) {
 			FullDownload(hfs, should_write_cache);
 		}
 		if (should_write_cache) {
-			current_cache->Insert(path, {length, last_modified, etag, version_id});
+			current_cache->Insert(path, GetCacheEntry());
 		}
 
 		if (!SkipBuffer()) {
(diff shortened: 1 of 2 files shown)


carlopi (changes requested):
Could you add a `require parquet` before the line of the test you added? That should fix CI failures

carlopi (approved):
Thanks! PR makes sense to me.

Just to make this clear, what's this is adding is that:
```
SET s3_secret_access_key='';
SET s3_region='us-east-2';
SET s3_endpoint='s3.amazonaws.com';
SET s3_access_key_id='';
FROM parquet_scan('s3://overturemaps-us-west-2/bridgefiles/2025-03-19.0-beta.0/dataset=Esri Community Maps/theme=buildings/type=building/part-00000-947c2149-23db-4abc-a7ee-3a892aac90ff.c000.snappy.parquet') LIMIT 1;
```
would otherwise fail.

Could it be isolated in a separate test? Basically the globbing has nothing to do with this, it's the fact that those 5 lines in current main will fail while they would succeed with this change.

carlopi:
I think PR it's good to go, thanks, but it could be handy to have a separate test for this.

PR #259, opened 2026-02-23
Make `server_id_pinnning` default to false

jeewonhh:
`server_id_pinning` introduced in https://github.com/duckdb/duckdb-httpfs/pull/248 was defaulting to `true`. I think the intent was to make it default to `false` (as per the PR description).


carlopi:
Thanks, I think here the PR description it's off, and it was a concious decision to have it as `true`.

Are you aware of issues deriving from this?

ywelsch:
We've noticed because we provide limited permissions on our DuckLake S3 buckets, and now DuckDB by default requires `GetObjectVersion` permission on a bucket to read from it.

carlopi:
I see, thanks for the feedback.

I did some basic search, and it seems that `x-amz-version-id` header is sent in any case for versioned buckets, so it seems the path to out out here it's overly complex, and then indeed the default should we switched back to default.

Then proposed change make sense to me, could you fix the test test/sql/copy/s3/version_id_pinning_defaults.test by opting in to the default?

Thanks!

jeewonhh:
Hey @carlopi , the test is now updated :)

carlopi:
@OneCyrus, fyi: after this PR (that double down as bug report), the default will go back to false, since it's same behaviour as before, and more testing will be needed to swap the default (and likely also a fallback / a way to specify this as a scoped secret property)

carlopi (approved):
Thanks!

carlopi:
And this will be in duckdb as part of https://github.com/duckdb/duckdb/pull/21062, thanks

PR #261, opened 2026-02-24
WIP: Fix incorrect hostname lookup error in curl backend

myrrc:
If curl backend for httpfs is enabled, requests to IPs with specific relative url paths like `//1.csv` fail in CURL with "couldn't resolve hostname" error.

There is more troubleshooting information available in https://github.com/vortex-data/vortex/pull/6627#issuecomment-3950388173, 
but in a nutshell, if you patch

```
diff --git a/src/httpfs_curl_client.cpp b/src/httpfs_curl_client.cpp
index 9d413e7..0e23432 100644
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -322,7 +322,7 @@ public:
 
                        // Add headers if any
                        curl_easy_setopt(*curl, CURLOPT_HTTPHEADER, curl_headers ? curl_headers.headers : nullptr);
-
+            curl_easy_setopt(*curl, CURLOPT_VERBOSE, 1L);
                        // Execute HEAD request
                        res = curl->Execute();
                        curl_easy_setopt(*curl, CURLOPT_NOBODY, 0L);
```

And launch

```
~/duckdb-httpfs ./build/debug/duckdb -c "SELECT * FROM 'http://127.0.0.1:8000//1.csv';"
* Could not resolve host: 1.csv
* shutting down connection #0
<...>
IO Error:
Could not resolve hostname error for HTTP HEAD to 'http://127.0.0.1:8000//1.csv'
~/duckdb-httpfs ./build/debug/duckdb -c "SELECT * FROM 'http://127.0.0.1:8000/1.csv';"
*   Trying 127.0.0.1:8000...
* connect to 127.0.0.1 port 8000 from 127.0.0.1 port 50582 failed: Connection refused
* Failed to connect to 127.0.0.1 port 8000 after 0 ms: Could not connect to server
* closing connection #0
*   Trying 127.0.0.1:8000...
* connect to 127.0.0.1 port 8000 from 127.0.0.1 port 50596 failed: Connection refused
* Failed to connect to 127.0.0.1 port 8000 after 0 ms: Could not connect to server
* closing connection #0
*   Trying 127.0.0.1:8000...
* connect to 127.0.0.1 port 8000 from 127.0.0.1 port 50602 failed: Connection refused
<...>
IO Error:
Could not connect to server error for HTTP HEAD to 'http://127.0.0.1:8000/1.csv'
```

[...]

PR #262, opened 2026-02-25
Fixup curl_url_set to take a CURLUPART_PATH, url baseline should already be provided

carlopi:
There is an issue where we were misusing the CURL API for building URL incrementally, and that would be visible when double slashes are present (since that's interpreted as a full file), this should fix that behaviour, I've validated locally, but waiting for CI results and reviews.

Testcase it's from https://github.com/duckdb/duckdb-httpfs/pull/261, that it's a PR adding the failing testcase and doubles also as issue report.

Thanks @myrrc!


carlopi:
CI validates this fixes the initial issue, but there is a failure that might be connected, I'll have a look in a bit.

PR #263, opened 2026-02-25
CURLOPT_ACCEPT_ENCODING set to NULL

carlopi:
Relevant documentation at https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html

I am considering whether having an option (that can be set user side), would be best here in case of regressions.

Fix was proposed by @ywelsch in the context of wider testing they performed where files uploaded to S3 like:
```
aws s3 cp file.json.gz s3://<bucket>/file.json.gz --content-type gzip
```
would be unzipped twice since information is not propagated between layers.

### DIFF
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -173,7 +173,7 @@ public:
 		// set connection timeout
 		curl_easy_setopt(*curl, CURLOPT_CONNECTTIMEOUT, http_params.timeout);
 		// accept content as-is (i.e no decompressing)
-		curl_easy_setopt(*curl, CURLOPT_ACCEPT_ENCODING, "identity");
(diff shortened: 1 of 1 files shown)


carlopi:
Relevant in the context of https://github.com/duckdb/duckdb/issues/20977

rustyconover:
Hi, this PR has caused some problems with me and I just opened an issue to re-enable cURL's support for Content-Encoding.  With this change no responses are compressed with gzip/deflate and we're using more bandwidth than we need.

But even worse, if I want to handle the decompression myself in an extension, now when the server sends `Content-Encoding: zstd` (because I added an `Accepts-Encoding` header), curl rejects the response with CURLE_BAD_CONTENT_ENCODING and the entire request fails.

rustyconover:
PR is #278

PR #265, opened 2026-02-27
Fix incorrect AWS list request in 1.4

myrrc:
If you use curl backend in v1.4, glob requests or generally everything involving S3 list request fails with something like

```
Other error: {"exception_type":"IO","exception_message":"URL using bad/illegal format or missing URL error for HTTP GET to '/?encoding-type=url&list-type=2&prefix=6685%2Fmerge%2F22445786768%2Ffineweb%2Fvortex-file-compressed%2F'"}
```

This doesn't happen with httplib backend.
The reason for this is that curl backend uses
`GetRequest->url` field for url propagation which is not set up correctly.

The flaw is actually in `BaseRequest(path, host)` constructor which initializes `url` to `path` only, ignoring `url`, but as it's in main DuckDB repo and the `url` is a `const string&` I can't change it here.

Fix this by manually concatenating host and path.

I couldn't reproduce this issue in v1.5, but there may be similar issues until `BaseRequest` is fixed.
I believe this fix should be backported to 1.4.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1244,8 +1244,13 @@ string AWSListObjectV2::Request(string &path, HTTPParams &http_params, S3AuthPar
 	// Get requests use fresh connection
 	string full_host = parsed_url.http_proto + parsed_url.host;
 	std::stringstream response;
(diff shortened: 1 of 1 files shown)


carlopi:
Thanks for your PR!

PR #272, opened 2026-03-03
Fix CURL client Post() not tracking total_bytes_received

### LINKED ISSUE #271: CURL client Post() method does not track total_bytes_received
## Problem

The CURL-based HTTP client's `Post()` method in `httpfs_curl_client.cpp` never updates `state->total_bytes_received` after executing a POST request. This causes the HTTPFS profiling output to always show `in: 0 bytes` when only POST requests are made.

For example, when using `EXPLAIN ANALYZE` on a query that communicates via HTTP POST, the profiling output shows:

```
┌─────────────────────────────────────┐
│┌───────────────────────────────────┐│
││         HTTPFS HTTP Stats         ││
││                                   ││
││            in: 0 bytes            ││
││            out: 9.7 KiB           ││
││              #HEAD: 0             ││
││              #GET: 0              ││
││              #PUT: 0              ││
││              #POST: 3             ││
││             #DELETE: 0            ││
│└───────────────────────────────────┘│
└─────────────────────────────────────┘
```

The `out:` bytes (sent) are tracked correctly, but `in:` bytes (received) are always zero for POST responses.

## Root Cause

In `httpfs_curl_client.cpp`, the `Get()` method correctly tracks received bytes:

```cpp
const idx_t bytes_received = request_info->body.size();
// ...
if (state) {
    state->total_bytes_received += bytes_received;
}
```

But the `Post()` method copies the response body to `info.buffer_out` and returns without ever incrementing `total_bytes_received`.

Note that the HTTPLib-based client (`httpfs_httplib_client.cpp`) correctly tracks POST response bytes via its `content_receiver` callback - only the CURL client is affected.

## Reproduction

Any workflow that uses HTTP POST requests and then checks the HTTPFS profiling stats will show `in: 0 bytes`.

rustyconover:
## Summary

- The CURL client's `Post()` method never updated `state->total_bytes_received`, causing HTTPFS profiling to always report `in: 0 bytes` for POST-only workloads
- Added the same `bytes_received` tracking pattern already used by `Get()` to `Post()`
- The HTTPLib client already tracks POST response bytes correctly via its `content_receiver` callback — only the CURL client was affected

## Test plan

- Run a workload that uses HTTP POST requests (e.g. an extension that communicates via POST)
- Use `EXPLAIN ANALYZE` to view the HTTPFS HTTP Stats profiling box
- Verify that `in:` now reflects the actual bytes received from POST responses instead of showing `0 bytes`

Fixes #271

### DIFF
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -417,6 +417,12 @@ public:
 
 		curl_easy_getinfo(*curl, CURLINFO_RESPONSE_CODE, &request_info->response_code);
 		info.buffer_out = request_info->body;
+
+		const idx_t bytes_received = request_info->body.size();
+		if (state) {
+			state->total_bytes_received += bytes_received;
+		}
+
 		// Construct HTTPResponse
 		return TransformResponseCurl(res);
 	}


carlopi:
Adding a test was intended? I think you could possibly add an `EXPLAIN ANALYZE` to some job uploading files to S3 (via the Minio.yml job)

carlopi:
I guess it sort of makes sense in any case, thanks!

PR #273, opened 2026-03-03
CURL: instead of using hard timeout use upload speed as proxy

carlopi:
With fixes suggested by @Mytherin (fixed time to parametrized timeout).


carlopi:
This solves a bug raised internally with overloaded networks.

PR #274, opened 2026-03-03
Format test/sql/copy/s3/version_id_pinning.test

carlopi:


carlopi:
This is trivial, test only

PR #278, opened 2026-03-07
Enable CURL response decompression

### LINKED ISSUE #277: Enable CURL response decompression
## Problem

The CURL client currently passes `NULL` to `CURLOPT_ACCEPT_ENCODING`, which means it does not advertise any supported compression encodings in the `Accept-Encoding` request header. As a result, servers never send compressed responses, leading to unnecessarily high bandwidth usage.

## Proposed Solution

Change `CURLOPT_ACCEPT_ENCODING` from `NULL` to `""` (empty string). Per the [CURL documentation](https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html), passing an empty string causes CURL to:

1. Advertise all compression encodings it supports (e.g. gzip, deflate, br, zstd) via the `Accept-Encoding` header
2. Automatically decompress responses that use those encodings

This is a one-line change that can significantly reduce bandwidth for HTTP transfers.

rustyconover:
## Summary

- Change `CURLOPT_ACCEPT_ENCODING` from `NULL` to `""` so CURL advertises all supported compression encodings (gzip, deflate, br, zstd) and automatically decompresses responses
- Reduces bandwidth usage for HTTP transfers with no behavioral change for callers

Fixes #277

## Test plan

- [ ] Verify CURL sends `Accept-Encoding` header with supported encodings
- [ ] Verify compressed responses are automatically decompressed
- [ ] Run existing httpfs test suite to confirm no regressions

### DIFF
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -185,8 +185,8 @@ public:
 		curl_easy_setopt(*curl, CURLOPT_TIMEOUT, http_params.timeout);
 		// set connection timeout
 		curl_easy_setopt(*curl, CURLOPT_CONNECTTIMEOUT, http_params.timeout);
-		// accept content as-is (i.e no decompressing)
-		curl_easy_setopt(*curl, CURLOPT_ACCEPT_ENCODING, NULL);
+		// advertise supported compression encodings and let CURL decompress responses automatically
+		curl_easy_setopt(*curl, CURLOPT_ACCEPT_ENCODING, "");
 		// follow redirects
 		curl_easy_setopt(*curl, CURLOPT_FOLLOWLOCATION, http_params.follow_location ? 1L : 0L);
 
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,6 +1,9 @@
 {
   "dependencies": [
     "openssl",
-    "curl"
+    {
+      "name": "curl",
+      "features": ["zstd"]
+    }
   ]
 }


carlopi:
Thanks for the PR, that would make sense, but before enabling this behaviour we would first need to fix duckdb/duckdb to handle receiving those pre-compressed streams.

Internal timeline is "nice for 1.5.1, but unclear"

rustyconover:
But curl decompresses it for you

carlopi:
I would be OK with adding an option, (that defaults to using NULL) but allows to opt in to "". Would that work for you?

The option we might need in any case to avoid regressions.
On defaulting to "" vs NULL, issue was that (as mentioned in the PR at https://github.com/duckdb/duckdb-httpfs/pull/263#issue-3990053997) uploading gzip-ped JSON files to S3 will have them doubly decompressed.
Given it's a niche but relevant data-source, I'd say first we need a fix before changing the default.

rustyconover:
Hey @carlopi,

You should let curl handle the decompression.

The whole argument around:

```aws s3 cp file.json.gz s3://<bucket>/file.json.gz --content-type gzip```

is flawed.  HTTP Compression is a content encoding for the transmission over the HTTP request, not a content-type.  The content type here should still be application/octet-stream not gzip.

I realize how it can be confusing for people, but we need to be pedantic about this.

I'd be happy to talk this through more. But I think this original motivation for #263 is flawed.

Rusty

PR #284, opened 2026-03-16
S3 globs: Faster globs for limited number of files OR limited numbers of files per folders

carlopi:
I introduced a regression in S3 glob performances via https://github.com/duckdb/duckdb-httpfs/pull/218.

This has been reported in multiple issues, such as https://github.com/duckdb/duckdb/issues/21385, https://github.com/duckdb/duckdb/issues/21347, and similar ones that correctly pointed out that `SET s3_allow_recursive_globbing = false;` would solve the issue.

Issue is basically that hierarchical / recursive globbing as a worse case penality equal to number of folders in the bucket, and given S3 default it's usually returning up to 1000 files, single request will be enough most of the times.

This PR refine the logic:
- perform a single listing (up to 1000 files)
   - no continuation? then we are completed with a single roundtrip, great
   - we have a continuation? check how many files have been returned AND how many unique folders. If there are num_folders < num_files / 100
       - hierarchical globs (first resetting the list)
       - regular listing based globs (from the continuation we already have)

This makes so for buckets with many files per folders (estimated on the first 1000 or so returned files) hierarchical strategy is used, since the worse case overhead that is linear on number of folders will be bounded and small.
If folders contain less than 100 files per folder, then use regular listing.

This should make so we have the best of both words:
* number of files in bucket / 1000 using lists
* 1 + number of folders + size of result set / 1000 using hierarchical listing

Also small buckets (less that 1000 files) are guaranteed to be done in 1 roundtrip.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -1064,6 +1064,7 @@ bool S3GlobResult::ExpandNextPath() const {
 	FileOpenerInfo info = {glob_pattern};
 	auto http_util = HTTPFSUtil::GetHTTPUtil(opener);
 	auto http_params = http_util->InitializeParameters(opener, info);
+	const vector<string> pattern_splits = StringUtil::Split(parsed_s3_url.key, "/");
 
 	vector<OpenFileInfo> s3_keys;
 	if (!current_common_prefix.empty()) {
@@ -1071,7 +1072,6 @@ bool S3GlobResult::ExpandNextPath() const {
 		auto prefix_path = parsed_s3_url.prefix + parsed_s3_url.bucket + '/' + current_common_prefix;
 
 		current_common_prefix = S3FileSystem::UrlDecode(current_common_prefix);
-		vector<string> pattern_splits = StringUtil::Split(parsed_s3_url.key, "/");
 		vector<string> key_splits = StringUtil::Split(current_common_prefix, "/");
 		const bool is_match =
 		    Match(key_splits.begin(), key_splits.end(), pattern_splits.begin(), pattern_splits.end(), false);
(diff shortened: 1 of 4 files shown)


Mytherin:
Cool - thanks for the PR!

This is a great start - but I think we might want to spend a bit more time thinking about the exact criteria for switching to the delimiter-based scan. 

For this delimiter based scan approach to speed things up you need selectivity on directories. Selectivity based on files is not relevant - after all, you're still scanning every matching directory fully. The strength in this scan method comes from being able to prune (ideally many?) directories. We might want to look at "directory-selectivity", instead of overall filter selectivity.

Another big component is the amount of files in each directory. If directories have few files in them, delimiter-based scans are slow. In the worst case of 1 file per directory we'll be doing one request per directory with this approach - which could result in very poor performance compared to scans, even if the filters are selective.

By doing a single request, there's also the "1001 file problem" - if we have processed 1000 files, clearly if there are 1001 files the fastest way of resolving the glob would be to follow the continuation token, regardless of anything else. 

I think getting these trade-offs right seems rather complex and a simple heuristic might not cut it. Given what we've seen about the potential for enormous performance degradations with this approach maybe it makes more sense to disable it by default for now, and spend some more time on optimizing this and investigating different scenarios before enabling it by default.

carlopi:
I should have edited the PR message as well, current logic it's based on file/directory indeed (basically, turned on only when there are an estimated number of > 100 files/directory). I think in that case it's easy to say the extra cost is negligible.

I am open to turning this off by default, but we should also consider the other angle: folder with lots of files / and possibly some selectivity, could be way worse, since we might do `O(total number of files)` vs `O(result-set) + O(number of folders)`.

src/s3fs.cpp
```diff
@@ -1110,11 +1113,57 @@ bool S3GlobResult::ExpandNextPath() const {
 			allow_s3_recursive_globbing = value.GetValue<bool>();
 		}
 
-		const bool use_recursive_glob = !StringUtil::Contains(parsed_s3_url.key, "**") && allow_s3_recursive_globbing;
+		const bool investigate_use_recursive_glob = !StringUtil::Contains(parsed_s3_url.key, "**") &&
```
Mytherin, line 1116: Maybe we should also check if there are any `non-*` filters in the folder list? We probably don't want to do this for `*/*.parquet` (I think?)

carlopi: I did some extra tests as part of https://github.com/duckdb/duckdb-httpfs/pull/286, I think current logic make sense to me, but there are a couple of TODO where we could do minor improvements.

Thanks for review and testing.

carlopi:
There is a follow up with more tests: https://github.com/duckdb/duckdb-httpfs/pull/286.

I am convinced this is a significant improvement over current state.

Compared to also turning the default, there are minor regressions we could look at improving when:
* there are small LIMIT applied (and still a lot of files / folder)
* situations like `*/*.parquet` that are basically very similar to `**` (not exactly, but)

PR #290, opened 2026-03-18
RemoveFiles: Avoid extra HEAD and avoid cutting last '/'

### LINKED ISSUE #766: ducklake_delete_orphaned_files fails with HTTP 404 on Cloudflare R2 (path-style S3)
### What happens?

When using DuckLake with Cloudflare R2 as the data storage backend and `s3_url_style = 'path'`, `ducklake_delete_orphaned_files` (and `CHECKPOINT`) fails with an HTTP 404 error.

## Setup

```sql
SET s3_url_style = 'path';
SET s3_endpoint = '<account>.r2.cloudflarestorage.com';
SET s3_region = 'auto';

ATTACH 'ducklake:postgres:postgresql://...' AS my_ducklake (
    DATA_PATH 's3://my-bucket/projects/my-project/',
    METADATA_SCHEMA 'my_schema',
    READ_WRITE
);
```

## Reproduce

```sql
CALL ducklake_delete_orphaned_files('my_ducklake');
-- or
CHECKPOINT;
```

## Error

```
HTTP Error: Failed to get files scheduled for deletion from DuckLake:
HTTP GET error reading
  's3://my-bucket/projects/my-project/?encoding-type=url&list-type=2&prefix=projects%2Fmy-project%2F'
  in region '' (HTTP 404 Not Found)
```

## Observations

The `DATA_PATH` is `s3://my-bucket/projects/my-project/`. Looking at the error URL, it appears that the sub-path `projects/my-project/` shows up both in the URL path and in the `prefix` query parameter. Our **guess** is that the ListObjectsV2 request is being constructed with a double prefix — the sub-path from `DATA_PATH` is included in the request URL *and* redundantly added as the `prefix` query parameter. We suspect the correct request should either list at the bucket root with the full prefix, or list at the data path without an additional prefix.

It's also possible the issue is related to R2 not supporting region auto-detection (`in region ''` in the error) or R2 handling path-style ListObjectsV2 requests differently from AWS S3.

Other maintenance functions (`flush_inlined_data`, `expire_snapshots`, `merge_adjacent_files`, `rewrite_data_files`, `cleanup_old_files`) all work fine with the same R2 setup. Only `delete_orphaned_files` (which needs to list S3 objects) fails.

### OS:

macos

### DuckDB Version:

1.3.x

### DuckLake Version:

0.3.x

### DuckDB Client:

python

### Hardware:

_No response_

### Full Name:

LittleLittleCloud

### Affiliation:

Individual

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have not tested with any build

No - Other reason (please specify in the issue body)

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

carlopi:
This should fix https://github.com/duckdb/ducklake/issues/766 properly AND avoid an unnecessary `HEAD` request in the `RemoveFiles` path.

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -841,7 +841,7 @@ void S3FileSystem::RemoveFiles(const vector<string> &paths, optional_ptr<FileOpe
 
 		const string &bucket = parsed_url.bucket;
 		if (keys_by_bucket.find(bucket) == keys_by_bucket.end()) {
-			string bucket_path = parsed_url.path.substr(0, parsed_url.path.length() - parsed_url.key.length() - 1);
+			string bucket_path = parsed_url.path.substr(0, parsed_url.path.length() - parsed_url.key.length());
 			if (bucket_path.empty()) {
 				bucket_path = "/";
 			}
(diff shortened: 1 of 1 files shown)


carlopi:
This does require proper testing on live S3 and R2 backends before merging.

PR #297, opened 2026-03-26
Fix util name mismatch

### LINKED ISSUE #296: Bug: http util naming mismatch
Hi team, if I don't read it wrong
- curl-based http util is named as [`HTTPFS-Curl`](https://github.com/duckdb/duckdb-httpfs/blob/de34f7b1f342071518d3903c556c846b48719487/src/httpfs_curl_client.cpp#L577)
- but when we switch http util, we use [`HTTPFSUtil-Curl`](https://github.com/duckdb/duckdb-httpfs/blob/de34f7b1f342071518d3903c556c846b48719487/src/httpfs_extension.cpp#L123) to decide whether to change
- It reads to me the mismatch will lead to new curl util creation every time

dentiny:
Closes https://github.com/duckdb/duckdb-httpfs/issues/296

As described in the issue, I think we should use the same naming at both ends, otherwise new curl http util will be created every single switch.

### DIFF
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -574,7 +574,7 @@ unordered_map<string, string> HTTPFSCurlUtil::ParseGetParameters(const string &t
 }
 
 string HTTPFSCurlUtil::GetName() const {
-	return "HTTPFSUtil-Curl";
+	return "HTTPFS-Curl";
 }
 
 } // namespace duckdb
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -574,7 +574,7 @@ unordered_map<string, string> HTTPFSCurlUtil::ParseGetParameters(const string &t
 }
 
 string HTTPFSCurlUtil::GetName() const {
-	return "HTTPFS-Curl";
+	return "HTTPFSUtil-Curl";
 }
 
 } // namespace duckdb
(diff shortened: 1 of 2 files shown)


carlopi:
Thanks, those string drifting was not intentional.

But now that we are there, wouldn't make sense to just keep the current logic? That is multiple:
```
SET httpfs_client_implementation = 'X';
```
must re-initialize the client (instead of being no-op) ?

What do you think?

carlopi:
And I meant:
avoiding the check on the name, and always reinitializing, basically removing matching on the name, that it's brittle (as demonstrated here).

Or maybe, if we want to check on the name, we could have it as static member, so that it's stored in a single place.

dentiny:
> avoiding the check on the name, and always reinitializing, basically removing matching on the name, that it's brittle (as demonstrated here).

To validate accessibility, there're two things to consider:
- whether the class is cheap to create and destruct (i.e., large memory consumption, connection pool)
- since we're storing shared pointer in the DB config, it could be used elsewhere, is there any issue to have multiple util impls in the DuckDB instance or process

I checked the existing HTTP util classes (base class, httplib impl, curl impl) and my multicurl-based extension, all of them are cheap to create and doesn't store any state, so I think it's fine to recreate every time.

I'm not too worried about the second item, so far it's a no-op change (compared with the current version).
When people (re)set http util config, they should update their util variable accordingly, if any.

> Or maybe, if we want to check on the name, we could have it as static member, so that it's stored in a single place.

I considered that -- but in theory we could have infinite number of http utils (i.e., I made my own in my extension) so I don't want to store official ones in a common header.

I will take the first approach. BTW, the minio test failure doesn't seem to be caused by my change?

carlopi:
Thanks!

PR #300, opened 2026-03-26
Fix secret type count passed to key-value reader

### LINKED ISSUE #299: BUG: miss handle secret type in S3 key-value reader
Hi team, I think for S3 filesystem we have 4 types of secrets, but somehow we only pass 3 of them into the key-value reader.
https://github.com/duckdb/duckdb-httpfs/blob/de34f7b1f342071518d3903c556c846b48719487/src/s3fs.cpp#L199-L200

dentiny:
Fixes https://github.com/duckdb/duckdb-httpfs/issues/299

I wish we could pass by `std::span`: https://en.cppreference.com/w/cpp/container/span.html

### DIFF
--- a/src/s3fs.cpp
+++ b/src/s3fs.cpp
@@ -197,7 +197,7 @@ S3AuthParams S3AuthParams::ReadFrom(optional_ptr<FileOpener> opener, FileOpenerI
 	}
 
 	const char *secret_types[] = {"s3", "r2", "gcs", "aws"};
-	S3KeyValueReader secret_reader(*opener, info, secret_types, 3);
+	S3KeyValueReader secret_reader(*opener, info, secret_types, sizeof(secret_types) / sizeof(secret_types[0]));
 
 	return ReadFrom(secret_reader, info.file_path);
 }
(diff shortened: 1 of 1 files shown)


src/s3fs.cpp
```diff
@@ -197,7 +197,7 @@ S3AuthParams S3AuthParams::ReadFrom(optional_ptr<FileOpener> opener, FileOpenerI
 	}
 
 	const char *secret_types[] = {"s3", "r2", "gcs", "aws"};
-	S3KeyValueReader secret_reader(*opener, info, secret_types, 3);
+	S3KeyValueReader secret_reader(*opener, info, secret_types, sizeof(secret_types) / sizeof(secret_types[0]));
```
carlopi:
```suggestion
	S3KeyValueReader secret_reader(*opener, info, secret_types, 4));
```

I think `S3KeyValueReader` needs a rework, ideally it should not be that low level to take const char ** + id.

I would say either fixing the bug in the simple way, or restructuring the `KeyValueReaders` interface, but I think simplest is just include `aws` as well.

carlopi (changes requested):
Thanks for finding this, I would keep it simpler and just put 4 for correctness, and move the refactor to later.

src/s3fs.cpp
dentiny: sure, if we decided to refactor I will use a plain 4 then -- updated!

carlopi (approved):

carlopi:
Thanks!

PR #305, opened 2026-04-07
Enable connection caching/reuse via `SET httpfs_connection_caching=true;`

carlopi:
Creating an Client to perform HTTP requests is somewhat heavy, and those can be cached across different usages.

Recycle of clients when their usage is done (but they are still valid) allows subsequent usages (in the same or different queries) to be faster.

Example of cases are like:
* sequential queries to the same API endpoint (like performing S3 globs)
* read of files in the same bucket (like `FROM read_parquet('s3://<fixed_bucket>/path/to/*.parquet');`
* frequent usage of common endpoints (also mixed)

Advantages can be from very visible (possibly say 3x in performing repeated HEAD requests) to somewhat visible, but more importantly this should be a strict improvement with no issues (if done right).

Given in case of any implementation problem, this might end up being very problematic, this is introduced opt-in via `SET httpfs_connection_caching=true;`. `RESET httpfs_connection_caching;` or `SET httpfs_connection_caching = false;` will restore the default state.

----

Implementation side, there is a global cache that act as a parcking lot of sort, on Curl HTTP client creation, if one is available on the same `base_url`, it moves to being used, freeing a spot. When a Curl HTTP client is not needed anymore (say for example destructor of FileHandle), those are returned to the cache/parking lot.

Logic to fill the parking spots is somewhat random:
* if any free spot, use that
* else, evict a random other client (that is bound to be older)

Note that in the cache/parking lot there can be multiple clients on the same `base_url`, since if those are used in parallel there might be multiple one that are useful.

Also, we add logging via `CALL enable_logging('HTTPFSInfo');`, a small example would be:
```sql
SET httpfs_connection_caching=true;
--- enable connection caching

CALL enable_logging('HTTPFSInfo');
FROM parquet_scan('s3://overturemaps-us-west-2/bridgefiles/2025-03-19.0-beta.0/dataset=Esri Community Maps/theme=buildings/type=building/part-00000-947c2149-23db-4abc-a7ee-3a892aac90ff.c000.snappy.parquet') LIMIT 1;
SELECT count(*) FROM duckdb_logs WHERE message LIKE '%connection_cache_miss%';
--- this should return some cache_miss

FROM parquet_scan('s3://overturemaps-us-west-2/bridgefiles/2025-03-19.0-beta.0/dataset=Esri Community Maps/theme=buildings/type=building/part-00000-947c2149-23db-4abc-a7ee-3a892aac90ff.c000.snappy.parquet') LIMIT 1;
SELECT count(*) FROM duckdb_logs WHERE message LIKE '%connection_cache_miss%';
--- this should return the same number of cache_miss, since they all come from the first round
```

Note that this cache only impact performances of requests, so this should be invisible at the SQL level, meaning the same physical request are performed.

### DIFF
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -15,10 +15,8 @@
 #include "duckdb/main/secret/secret_manager.hpp"
 #include "http_state.hpp"
 
-#include <chrono>
 #include <map>
 #include <string>
-#include <thread>
 
 #include "s3fs.hpp"
 
@@ -557,7 +555,7 @@ bool HTTPFileSystem::ReadInternal(FileHandle &handle, void *buffer, int64_t nr_b
 		DUCKDB_LOG_FILE_SYSTEM_READ(handle, nr_bytes, location);
 		// Update handle status within critical section for parallel access.
 		if (hfh.flags.RequireParallelAccess()) {
-			std::lock_guard<std::mutex> lck(hfh.mu);
+			std::lock_guard<mutex> lck(hfh.mu);
 			hfh.buffer_available = 0;
 			hfh.buffer_idx = 0;
 			hfh.file_offset = location + nr_bytes;
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -1029,6 +1029,10 @@ HTTPFileHandle::~HTTPFileHandle() {
 	}
 }
 
+void HTTPFSUtil::ClearCachedConnections() {
+	// no-op by default
+}
+
 string HTTPFSUtil::GetName() const {
 	return "HTTPFS";
 }
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -1011,9 +1011,9 @@ unique_ptr<HTTPClient> HTTPFileHandle::GetClient() {
 }
 
 unique_ptr<HTTPClient> HTTPFileHandle::CreateClient() {
-	// Create a new client
 	string path_out, proto_host_port;
 	HTTPUtil::DecomposeURL(path, path_out, proto_host_port);
+	cached_proto_host_port = proto_host_port;
 	return http_params.http_util.InitializeClient(http_params, proto_host_port);
 }
 
@@ -1023,6 +1023,18 @@ void HTTPFileHandle::StoreClient(unique_ptr<HTTPClient> client) {
 
 HTTPFileHandle::~HTTPFileHandle() {
 	DUCKDB_LOG_FILE_SYSTEM_CLOSE((*this));
+	if (!cached_proto_host_port.empty()) {
+		auto &httpfs_util = static_cast<HTTPFSUtil &>(http_params.http_util);
+		auto client = client_cache.GetClient();
+		while (client) {
+			httpfs_util.CloseClient(cached_proto_host_port, std::move(client));
+			client = client_cache.GetClient();
+		}
+	}
+}
+
+void HTTPFSUtil::CloseClient(const string &, unique_ptr<HTTPClient> &&) {
+	// no-op by default, client destroyed via RAII
 }
 
 string HTTPFSUtil::GetName() const {
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -1013,7 +1013,6 @@ unique_ptr<HTTPClient> HTTPFileHandle::GetClient() {
 unique_ptr<HTTPClient> HTTPFileHandle::CreateClient() {
 	string path_out, proto_host_port;
 	HTTPUtil::DecomposeURL(path, path_out, proto_host_port);
-	cached_proto_host_port = proto_host_port;
 	return http_params.http_util.InitializeClient(http_params, proto_host_port);
 }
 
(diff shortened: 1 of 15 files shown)


src/httpfs_connection_caching_client.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+namespace duckdb {
+
+unique_ptr<HTTPClient> HTTPFSCachedUtil::FindCachedCandidate(const string &proto_host_port) {
+	if (proto_host_port.empty()) {
+		return nullptr;
+	}
+	if (auto lock = std::unique_lock<std::mutex>(cached_httpclients_mutex, std::try_to_lock)) {
+		for (idx_t i = 0; i < cached_httpclients.size(); i++) {
+			if (cached_httpclients[i].proto_host_port == proto_host_port && cached_httpclients[i].cached_client) {
```
Mytherin: Maybe also having an unordered_map would be better?

src/include/httpfs_client.hpp
```diff
@@ -58,6 +81,21 @@ class HTTPFSCurlUtil : public HTTPFSUtil {
...
+	unique_ptr<HTTPClient> InitializeClient(HTTPParams &http_params, const string &proto_host_port) override;
+	void CloseClient(const string &proto_host_port, unique_ptr<HTTPClient> &&client) override;
+	unique_ptr<HTTPResponse> SendRequest(BaseRequest &request, unique_ptr<HTTPClient> &client) override;
+
+	string GetName() const override;
+
+	bool EnableCaching(BaseRequest &request);
+	unique_ptr<HTTPClient> FindCachedCandidate(const string &proto_host_port);
+	void StoreCachedCandidate(const string &proto_host_port, unique_ptr<HTTPClient> &&client);
+	std::mutex cached_httpclients_mutex {};
```
Mytherin: `mutex` as a public variable as a rule is not a great idea - very error prone. Can we move cached clients to a separate `HTTPCache` class that isolates all of this?

src/httpfs.cpp
```diff
@@ -1023,6 +1023,18 @@ void HTTPFileHandle::StoreClient(unique_ptr<HTTPClient> client) {
 
 HTTPFileHandle::~HTTPFileHandle() {
 	DUCKDB_LOG_FILE_SYSTEM_CLOSE((*this));
+	if (!cached_proto_host_port.empty()) {
+		auto &httpfs_util = static_cast<HTTPFSUtil &>(http_params.http_util);
+		auto client = client_cache.GetClient();
+		while (client) {
+			httpfs_util.CloseClient(cached_proto_host_port, std::move(client));
```
Mytherin: Maybe `cached_proto_host_port` should be moved into the client - given that it is critical to know when a client can be (re)used?

src/httpfs_connection_caching_client.cpp
```diff
@@ -0,0 +1,135 @@
...
+			if (!cached_httpclients[i].cached_client) {
+				cached_httpclients[i].cached_client = std::move(client);
+				cached_httpclients[i].proto_host_port = proto_host_port;
+				return;
+			}
+		}
+		// Else evict one at random
+		{
+			RandomEngine engine;
+			size_t index = engine.NextRandomInteger() % cached_httpclients.size();
```
Mytherin: Maybe an LRU is better here

Mytherin (commented):
Thanks! Very cool feature - left some comments

src/httpfs_connection_caching_client.cpp
carlopi: I think it would need to be a multimap, or unordered_map <string, vector< > >, but then I think it's simpler to just iterate a bounded vector (we have max 64 slots).

Also, loop is now like:
```c++
                for (auto &entry : entries) {
                        if (entry && entry->GetBaseUrl() == base_url) {
                                return std::move(entry);
                        }
                }
```
that should be clean enough.

src/include/httpfs_client.hpp
carlopi: Thanks, moved to a helper class that manages the Cache.

src/httpfs.cpp
carlopi: Done in https://github.com/duckdb/duckdb/pull/21924

src/httpfs_connection_caching_client.cpp
carlopi: I can see this is hacky, one thing where LRU do not maps perfectly is that this is a sort of parking lot, so it's not obvious what it means last used, meaning last returned would benefit code that holds to clients for too long.

I think random is as good as other strategies, maybe a bit heavy on random number generator.
For now current implementation it's still random based, let me look at what else could make sense.

src/include/httpfs_client.hpp
```diff
@@ -35,12 +53,26 @@ struct HTTPFSParams : public HTTPParams {
...
 };
 
+class HTTPClientConnectionCache {
+public:
+	unique_ptr<HTTPClient> Find(const string &base_url);
+	void Store(unique_ptr<HTTPClient> &&client);
+	void Clear();
+
+private:
+	std::mutex mutex {};
```
Mytherin: nit: let's avoid using `std::`

PR #307, opened 2026-04-09
Fix test/sql/connection_caching.test

carlopi:


carlopi:
Test only, should be irrelevant.

PR #308, opened 2026-04-14
fix: avoid ODR-use of static constexpr LEVEL in HTTPFSInfoLogType

qsliu2017:
The constructor passed LEVEL by const-reference to LogType(), which is an ODR-use requiring an out-of-line definition. In Debug builds (-O0) the constructor is not inlined, so the linker fails with "undefined reference to HTTPFSInfoLogType::LEVEL" when linking plan_serializer.

Use the enum literal directly to avoid the ODR-use.

Related CI: https://github.com/duckdb/pg_duckdb/actions/runs/24387367520/job/71224352195?pr=1025


carlopi:
Thanks!!

PR #313, opened 2026-04-20
fix: use stoull instead of stoi/stoll for Content-Length parsing

benben:
std::stoi returns int (max 2,147,483,647). When downloading files larger than 2GB, the Content-Length HTTP header value overflows int and throws std::out_of_range. DuckDB surfaces this as "Invalid Error: stoi", crashing the query.

This affects any read_parquet() call on files >2GB served over HTTP/S3.

Fix: replace std::stoi with std::stoull (unsigned 64-bit) in both call sites that parse Content-Length headers. Also wrap in try-catch for robustness against malformed headers.

### DIFF
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -337,13 +337,16 @@ unique_ptr<HTTPResponse> HTTPFileSystem::GetRangeRequest(FileHandle &handle, str
 			    }
 
 			    if (response.HasHeader("Content-Length")) {
+				    unsigned long long content_length;
+				    bool parsed = false;
 				    try {
-					    auto content_length = stoull(response.GetHeaderValue("Content-Length"));
-					    if ((idx_t)content_length != buffer_out_len) {
-						    RangeRequestNotSupportedException::Throw();
-					    }
+					    content_length = stoull(response.GetHeaderValue("Content-Length"));
+					    parsed = true;
 				    } catch (const std::exception &) {
 					    // Content-Length header contains a non-numeric value — skip validation.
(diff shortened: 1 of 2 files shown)


src/httpfs.cpp
```diff
@@ -335,9 +337,13 @@ unique_ptr<HTTPResponse> HTTPFileSystem::GetRangeRequest(FileHandle &handle, str
...
-				    if ((idx_t)content_length != buffer_out_len) {
-					    RangeRequestNotSupportedException::Throw();
+				    try {
+					    auto content_length = stoull(response.GetHeaderValue("Content-Length"));
+					    if ((idx_t)content_length != buffer_out_len) {
+						    RangeRequestNotSupportedException::Throw();
+					    }
+				    } catch (const std::exception &) {
+					    // Content-Length header contains a non-numeric value — skip validation.
 				    }
```
carlopi, line 345: Thanks, but in cases we throw via `RangeRequestNotSupportedException::Throw`, we would like that to pass through, not be catched right away.

Maybe try { } catch (...) { } only around stoll / stoull ?

carlopi (changes requested):

src/httpfs.cpp
benben: Hi @carlopi! Thanks for the feedback! I've updated the branch.

carlopi (approved):
Thanks!

benben:
@carlopi please merge 😊

carlopi:
Thanks!

PR #315, opened 2026-04-23
Replace dynamic_cast with name check and static_cast in connection caching callback

DinosL:
Fixes: https://github.com/duckdblabs/duckdb-internal/issues/8922
Fixes: https://github.com/duckdblabs/madhive/issues/17


carlopi:
Thanks!

Once this is built it should be available (for v1.5.2) at a nightly versioned endpoint, like `FORCE INSTALL httpfs FROM core_nigthly VERSION 'xyz';`
I think it would be handy verifying if doing so works around to both issues.

PR #318, opened 2026-04-24
feat(gcs): add credential_chain provider using ADC

marhar:
## Summary

Adds a `credential_chain` provider for the `gcs` secret type that obtains an OAuth2 access token via Google Application Default Credentials, mirroring the convenience of the existing `aws` `credential_chain` provider.

```sql
CREATE SECRET (TYPE gcs, PROVIDER credential_chain);
```

Two ADC sources are supported, in the standard order:

1. **gcloud user credentials** — reads `~/.config/gcloud/application_default_credentials.json` and exchanges the refresh token at `oauth2.googleapis.com/token` for a 1h access token.
2. **GCE metadata server** — `http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token` with the required `Metadata-Flavor: Google` header and a 2s timeout so non-GCE hosts fail fast.

Service-account JSON keys (require JWT signing) and workload identity federation are not yet implemented and can be added in follow-up PRs.

## Notes for reviewers

- JSON parsing uses crude string extraction. This is sufficient for the well-formed output of `gcloud` and the Google token endpoint, but should ideally be replaced with `yyjson` before merging — happy to do that in this PR if preferred.
- The created secret sets `refresh = auto` so the existing httpfs refresh hook (`s3fs.cpp:737`) re-invokes this provider on a 401 from a stale token. This handles persistent secrets reused across sessions and long-running sessions opening new files post-expiry. It does *not* handle a mid-stream 401 during a single read, nor proactive refresh.
- Source 2 (gcloud) is silent when the ADC file is absent; source 4 (metadata) is silent on network/DNS errors so non-GCE hosts don't see noise. Errors from sources that *were* attempted are surfaced together if the chain ultimately yields no token.

## Test plan

- [ ] Local: `gcloud auth application-default login` → `CREATE SECRET (TYPE gcs, PROVIDER credential_chain);` → `SELECT * FROM 'gs://<bucket>/<file>';`
- [ ] GCE/GKE instance with attached service account: same `CREATE SECRET` → query a `gs://` URL
- [ ] Refresh path: persistent secret, wait >1h, open a new file in a new session — verify the 401 → refresh path produces a fresh token
- [ ] Negative: no ADC file and not on GCE → `CREATE SECRET` raises a clear `IOException` listing the sources tried

### DIFF
--- a/src/create_secret_functions.cpp
+++ b/src/create_secret_functions.cpp
@@ -2,6 +2,12 @@
 #include "s3fs.hpp"
 #include "duckdb/main/extension/extension_loader.hpp"
 #include "duckdb/common/local_file_system.hpp"
+#include "duckdb/common/http_util.hpp"
+#include "duckdb/main/database.hpp"
+#include "duckdb/main/client_context.hpp"
+
+#include <cctype>
+#include <cstdio>
 
 namespace duckdb {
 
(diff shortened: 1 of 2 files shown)


marhar:
Closing — opened on the wrong repo. Moving to marhar/duckdb-httpfs as a reference PR.

PR #321, opened 2026-05-05
fix: Fix ASAN issue

gnusi:
Fix UB caused by passing a temporary `std::string` (url.endpoint + fragment_next_page_url) directly into GetRequestInfo.

I'd strongly encourage to store a plain `std::string` instead of `const std::string&` in `BaseRequestInfo` to avoid such issues in future. If you'd like to go this way I can contribute, but that'd be a bigger change and out of scope of this PR.


lnkuiper:
Superseded by https://github.com/duckdb/duckdb/pull/22953

PR #329, opened 2026-05-22
Removed local tests in test_csv_remote.test

sebastiaan-dev:
Previous path change did not fix the issue, as duckdb submodule is not available on the same path when running CI in the duckdb repository.

Removed the tests now, similar to: https://github.com/duckdb/duckdb-httpfs/pull/215


carlopi:
I am not sure this change makes complete sense, what the test is testing is that local access path and remote access path have the same results.

I am not sure what's the proper solution, possibly the DUCKDB path needs to be passed externally?

sebastiaan-dev:
@carlopi I'm not sure how this differs from the change in https://github.com/duckdb/duckdb-httpfs/pull/215? I looked at mapping a different path, but then came across this PR which deleted it on main.

carlopi:
Right, indeed, thanks!

PR #330, opened 2026-05-22
Init DRBG single-threadedly to fix data race

smvv:
- Init DRBG once to fix TSAN data race in crypto. See https://github.com/duckdb/duckdb/pull/22833.

### DIFF
--- a/src/include/crypto.hpp
+++ b/src/include/crypto.hpp
@@ -4,6 +4,7 @@
 #include "duckdb/common/encryption_functions.hpp"
 #include "duckdb/common/helper.hpp"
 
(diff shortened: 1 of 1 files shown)


carlopi:
I think this PR is doing multiple things, logically I think:
1. TSAN data race
2. namespace runners
3. move to main

are separated PRs / concerns.

smvv:
ok i split the PR. This is only TSAN and the rest in https://github.com/duckdb/duckdb-httpfs/pull/331.

It's not so easy to separate the runners PR from the main PR because runners requires the main branch for CI tools.

carlopi:
Thanks!

PR #332, opened 2026-05-24
Add credentials refresh in more locations

mst:
Supersedes #165.

Rebased `auto_refresh_fix` onto current `main`.

TODO follow-ups:
- [ ] batched `RemoveFiles` refresh support
- [ ] 4xx handling for non-GET paths
- [ ] regression test
- [ ] Thread safety / single-flight refresh

### DIFF
--- a/scripts/test_refresh_get_range.sh
+++ b/scripts/test_refresh_get_range.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+# Regression test for PR #165 -- ExecuteWithRefresh on S3FileSystem::GetRangeRequest.
+#
+# The existing refresh hook lives in S3FileHandle::Initialize and only fires
+# on HEAD failures. Callers that skip HEAD (DuckLake, Iceberg, anything using
+# prefilled OpenFileInfo.extended_info) silently opt out of auto-refresh.
+# We trigger the same HEAD-skip via enable_http_metadata_cache, then revoke
+# the service-account token mid-session so phase B's first range GET hits a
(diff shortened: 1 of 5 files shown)


KumoSiunaus:
I built `auto_refresh_fix` against a local MinIO and can confirm it fixes refresh on the glob-listing and range-GET paths. To help with the `regression test` and `4xx handling for non-GET paths` TODOs, I opened a PR into your branch: https://github.com/mst/duckdb-httpfs/pull/1

It adds (1) a deterministic sqllogictest covering the open/glob/range-GET/write paths (passes on your branch, fails on `main` at the new surfaces), and (2) a small fix so the write path actually refreshes — `Put`/`PostRequest` report a 4xx via the response status rather than throwing, so the error was raised outside `ExecuteWithRefresh` and writes never refreshed.

One thing to flag for the `Thread safety / single-flight refresh` TODO: because `ExecuteWithRefresh` mutates a shared `auth_params` by reference, concurrent range GETs (parallel scans) or multipart part PUTs can race on the credential fields during a refresh. Details in the PR.

KumoSiunaus:
Follow-up: I've added the single-flight / thread-safety piece for the shared `auth_params` (one refresh per expiry via a generation counter + snapshot-under-lock for signing, covering the parallel range-GET and multipart part-PUT paths) to that PR — https://github.com/mst/duckdb-httpfs/pull/1. Addresses the "single-flight refresh" TODO; happy to reshape it however fits best.

lnkuiper:
This has been superseded by work already merged into `main`

PR #336, opened 2026-06-03
Change curl timeout to per-read instead of total connection time

DinosL:
TL;DR: `curl` enforces `http_timeout` as a total transfer time, while `httplib` per-read.

Longer explanation here: https://github.com/duckdblabs/duckdb-internal/issues/9552#issuecomment-4610632082


carlopi:
This looks great, sorry I didn't came back on this earlier, retargeted so it can ship with `v1.5.5`.

carlopi (approved):
Thanks!

Issue #282, opened 2026-03-11
curl backend doesn't use system certificates on MacOS


freundTech:
I have duckdb installed from homebrew on MacOS. I just installed the 1.5.0 update, which enabled the `curl` backend by default.
We use the `httpfs` extension for accessing S3 compatible object stores with self-signed certificates. The certificates are installed to the MacOS certificate store.

The old `httplib` backend would automatically trust those certificates, while the new `curl` backend fails with
```
IO Error:
SSL peer certificate or SSH remote key was not OK error for HTTP HEAD to 'https://***'
```

Linux is not affected by this.

It's possible to work around this problem by switching back to the `httplib` backend using `SET httpfs_client_implementation=httplib;`

carlopi:
Thanks for raising this bug.

We should autodetect the certificate path, but apparently something it's off in your case.

I would recommend (both as a workaround AND as a way to get more informations), the following:

 ```
curl -v https://duckdb.org/data/latest_stable_version.txt
```
This will use native curl to reach https://duckdb.org/data/latest_stable_version.txt, and it should log in the first 20 lines something like:
```
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
*  CAfile: /Users/.../path/to/.../cacert.pem
*  CApath: none
```

then, duckdb side, you could try running the following SQL:
```sql
SET ca_cert_file = '/Users/.../path/to/.../cacert.pem';
FROM read_blob('https://duckdb.org/data/latest_stable_version.txt');
```

If this works, it would mean the issue is in identifying the path to `cacert.pem` and offer a workaround.
If not, that's unexpected but relevant information.

This will not solve the bug, auto-detection should just work in most/all settings, but provide a path forward.

freundTech:
`curl` on MacOS doesn't use a CAfile. It instead uses the Apple SecTrust API directly:

```
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* SSL Trust Anchors:
*   Native: Apple SecTrust
*   OpenSSL default paths (fallback)
```

(note that this is homebrew installed `curl`. MacOS preinstalled `curl` lies about using `/etc/ssl/cert.perm` but actually still uses Apple SecTrust API. When telling homebrew `curl` to use that with `--no-ca-native --cacert /etc/ssl/cert.perm` requests to self-signed sites fail because it doesn't include the self-signed certificates).

carlopi:
I am not sure if that's true in general, for me:
```
[...]

PR #337, opened 2026-06-03
Don't set CURL_NATIVE_CA if a cert path is provided

freundTech:
I'm working on exposing Apple SecTrust support in the `curl` vcpkg package (https://github.com/microsoft/vcpkg/pull/52150, https://github.com/microsoft/vcpkg/pull/52158) to make it possible to solve #282. Depending on what the vcpkg maintainers decide this will be either an optional feature for duckdb-httpfs to enable or just enabled by default.

Testing with this enabled surfaced a small bug in duckdb-httpfs: Settign `SET ca_cert_file = 'certificate_file';` will still fall-back to the native CA store, if one is available, instead of erroring when the certificate can't be validated using the provided cert file, causing `test/sql/httpfs/ca_cert_file.test` to fail.

This commit fixes this behavior to prepare for adding SecTrust support and fixing #282 later.


freundTech:
The VCPKG maintainers decided to merge the version that always enables the curl build option, because they expect most users to want that behavior by default.

This means that without this fix the duckdb-httpfs test suite will likely fail when built with the new vcpkg build manifest.

PR #339, opened 2026-06-11
Fix duration calculation

dentiny:
Closes https://github.com/duckdblabs/duckdb-internal/issues/8877

Current implementation uses system clock, which doesn't guarantee monotonic increase which leads to numerically negative duration and cast failure.
https://github.com/duckdb/duckdb/blob/73a804ef72ea1d1935b12c6432bd634887fe91ea/src/common/types/timestamp.cpp#L445-L449

### DIFF
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -489,7 +489,7 @@ bool HTTPFileSystem::TryRangeRequest(FileHandle &handle, string url, HTTPHeaders
                                      char *buffer_out, idx_t buffer_out_len) {
 	auto &hfh = handle.Cast<HTTPFileHandle>();
 
-	const auto timestamp_before = Timestamp::GetCurrentTimestamp();
+	const auto timestamp_before = std::chrono::steady_clock::now();
 	auto res = GetRangeRequest(handle, url, header_map, file_offset, buffer_out, buffer_out_len);
 
 	if (res) {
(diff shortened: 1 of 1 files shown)


carlopi:
Would this not need to fix `GetCurrentTimestamp` in general? Or maybe keep using that bup cap to 0 (when negative)?

dentiny:
> Would this not need to fix `GetCurrentTimestamp` in general?

TBH I don't think there's anything wrong to return system clock timestamp for the util function, which could be used, say, to log the current human-readable timestamp.
It's a behavior change, so we need to make sure it doesn't affect behavior for all existing call sites.

For code consistency, it'd be nicer if we could add a counterpart time util function to steady block util. 
(I didn't do that for now, because it involves DuckDB core side change, and I'd love to see the fix be in v1.5.3 release if possible, I could do proper things as a followup later this week)

> Or maybe keep using that bup cap to 0 (when negative)?

It's a workaround not a root cause fix -- we should report the real duration.

dentiny:
> 1. test/sql/copy/csv/test_url_with_plus.test:10
Query unexpectedly failed! (test/sql/copy/csv/test_url_with_plus.test:10)!
FROM read_csv('https://d37ci6vzurychx.cloudfront.net/misc/taxi+_zone_lookup.csv');
HTTP Error: HTTP GET error on 'https://d37ci6vzurychx.cloudfront.net/misc/taxi+_zone_lookup.csv' (HTTP 0 Internal Server Error)

Failed minio test is not related to my PR.

carlopi:
Thanks!!

PR #340, opened 2026-06-12
Add s3_compatible_url_schemes setting to route additional URL schemes to the S3 filesystem

plusplusjiajia:
## Motivation

  Many object stores are S3-compatible but use their own URL scheme (e.g. `oss://`, `cos://`, `obs://`). Data lake catalogs (e.g. Iceberg REST catalogs) return table locations using these schemes, which the client cannot control — DuckDB currently fails with "No such file or directory"
  because no filesystem recognizes the prefix, even though the S3 filesystem could serve these requests via a custom endpoint.

  httpfs already handles this pattern for specific vendors with the built-in `r2://` and `gcs://`/`gs://` schemes. This PR generalizes that into a setting instead of hardcoding more vendor schemes one by one.

  ## Changes

  - New `s3_compatible_url_schemes` setting (VARCHAR, comma-separated): registered schemes are recognized by `S3FileSystem::TryGetPrefix` / `CanHandleFile` and included in the default scope of newly created S3 secrets.
  - The setting is **global-only** (`default_scope = GLOBAL`; explicit `SET SESSION` is rejected) — URL-to-filesystem routing is instance-level by nature.
  - Schemes of built-in filesystems (`s3://`, `gcs://`, `r2://`, `http://`, `https://`, `hf://`, `file://`, ...) are rejected to prevent hijacking their routing.
  - `RESET` restores the default (empty) behavior; covered by tests.

  Default behavior is unchanged: with the setting unset, routing and secret scopes are identical to before.

  ## Usage

  ```sql
  SET s3_compatible_url_schemes = 'oss';
  CREATE SECRET (TYPE S3, KEY_ID '...', SECRET '...', ENDPOINT 'oss-cn-hangzhou.aliyuncs.com');
  SELECT * FROM 'oss://my-bucket/path/file.parquet';

  Notes

  - Considered a per-instance design (threading the scheme list through S3AuthParams), but CanHandleFile/TryGetPrefix are static with no context available; the process-level registry with global-only SET keeps semantics consistent and the change minimal.
  - The settings list in duckdb-web should get a new entry once this lands.


plusplusjiajia:
Hi @carlopi, could you trigger CI when you get a chance? This generalizes the hardcoded r2:///gcs:// scheme handling into a s3_compatible_url_schemes setting, so S3-compatible stores like oss:// work via their native scheme. No behavior change when unset; tests included. Thanks!

plusplusjiajia:
Hi @lnkuiper — rebased onto your S3Provider refactor; the setting now just extends S3Provider. Could you approve the workflow run? Happy to address any feedback

lnkuiper:
Thanks for the PR! The overall direction makes sense, but I think the setting and its ownership need a little more work before this is safe to merge. Here's my (Codex-assisted) review:

1. Could this be a `LIST(VARCHAR)` setting, perhaps named `s3_url_scheme_aliases` or `s3_additional_url_schemes`? Extension settings support `LogicalType::LIST(LogicalType::VARCHAR)`, so the interface could be:

   ```sql
   SET GLOBAL s3_url_scheme_aliases = ['oss', 'cos'];
   ```

   This supports multiple schemes without comma parsing or escaping rules and follows existing list settings such as `allowed_directories`. I would accept one representation (preferably bare scheme names), lowercase it, deduplicate it, reject `NULL` elements, and validate the scheme syntax. The current string handling turns `NULL` into `null://` and `oss://bucket` into `oss://bucket://`.

2. `GLOBAL` seems like the correct SQL scope, but process-global static state is not. Filesystem dispatch calls `CanHandleFile(path)` without a `FileOpener`/`ClientContext`, and the same database may have several client contexts, so a session-local value could not reliably control routing. I also reproduced the current static state leaking between two independent `DatabaseInstance`s: changing the setting in one database changed routing in the other while `current_setting()` still showed different values.

   I would keep the setting value in `DBConfig`, which already owns global extension settings, rather than add a new field directly to `DatabaseInstance` or put it in `ClientContextState`. `S3FileSystem` is registered once per database and already reaches that database through its `BufferManager`, so `CanHandleFile` can read the database's setting. If that lookup needs caching, the cache should be per `S3FileSystem`/database and use an immutable snapshot. The callback should only validate/normalize the list; it should not maintain a second process-global copy.

   The static `S3Provider`/`S3Url` helpers currently assume context-free matching. I think the configured aliases (or an already resolved provider match) should be passed into that path from `S3FileSystem`/the available `FileOpener`. Secret creation already receives a `ClientContext`, so it can read the same global setting from the database when constructing scopes.

3. Multiple routing aliases also need defined secret behavior. At present every newly created S3 secret gets every configured alias in its default scope. With separate `oss_secret` and `cos_secret` secrets, `which_secret` selected the same secret for both URLs because the scopes tied. I think custom aliases should require an explicit secret `SCOPE`, unless the setting models a scheme-to-secret association. Please add a test with two aliases and two differently scoped secrets.

4. `reserved_schemes` duplicates `ProviderMatches()` in this file and DuckDB core's known extension prefixes, while still missing arbitrary registered filesystems. Could we avoid maintaining another list? At minimum, derive the httpfs-owned S3/GCS/R2 schemes from `ProviderMatches()`. If conflicts with all registered filesystems must be rejected, that check likely belongs in or needs support from the VFS/core registry.

5. Please add a two-database isolation regression test, list/validation tests, and the two-secret test above. The test should avoid a live request to Aliyun; a local parse/routing failure such as `oss:///test.csv` is enough to prove that the S3 filesystem handled the path. The autoload-registry entry should land with this change as well so the setting does not require an explicit `LOAD httpfs`.

plusplusjiajia:
@lnkuiper  Thanks for the detailed review — the two-database leak and the null:// / oss://bucket:// cases were all real.

  1. LIST(VARCHAR), bare names, lowercased, deduplicated, NULL and bad syntax rejected.
  2. No process-global state — the callback only validates, routing reads the value from the owning database's DBConfig. C++ two-database regression test included.
  3. Aliases no longer join the default secret scope; two-alias/two-secret which_secret test added.
  4. Reserved schemes derive from ProviderMatches() + core's EXTENSION_FILE_PREFIXES. Two schemes that table misses are still hardcoded: file (aliasing it hijacks every file:// url — the VFS only falls back to LocalFileSystem when nothing claims a path) and abfs (duckdb-azure serves it; only abfss:// is listed). Both look worth keeping to me; the cleaner fix is adding them to the core table, which I'm happy to do in a follow-up if you prefer.
  5. Tests are local now. Autoload-registry PR coming; the explicit LOAD httpfs stays until this repo's pin has it.

lnkuiper:
One correctness issue remains: endpoint validation happens before URL parameters are applied.

`S3Provider::ReadAuthParams()` reads secrets/settings and immediately calls `InitializeAuthParams()`. For an aliased URL without an endpoint in those sources, the new guard throws before `S3Url::Resolve()` can read `?s3_endpoint=...`. R2 has the same ordering problem, even though its error says the endpoint may come from the URL.

Could we handle this through a shared finalization step instead of special-casing aliases? I think the flow should be:

1. Read secrets/settings and apply provisional provider defaults. A missing required endpoint for an alias or R2 should remain unresolved here and must not fall back to AWS.
2. Read URL query parameters, which have the final opportunity to supply or override the endpoint.
3. Call something like `FinalizeAuthParams()` to validate that aliases/R2 have a non-empty endpoint and then apply the final derived defaults.

This should preserve all of these cases:

- endpoints from a secret, `s3_endpoint`, or `?s3_endpoint=...` succeed;
- an endpoint missing from all sources fails;
- `?s3_endpoint=` or whitespace still fails after clearing an earlier value;
- aliases never silently receive the AWS endpoint default.

Please add coverage that passes an alias and R2 endpoint only through the URL, plus the existing cases where a URL parameter clears an endpoint supplied by a secret or setting.

Ready to merge once this is fixed.

plusplusjiajia:
> One correctness issue remains: endpoint validation happens before URL parameters are applied.
> 
> `S3Provider::ReadAuthParams()` reads secrets/settings and immediately calls `InitializeAuthParams()`. For an aliased URL without an endpoint in those sources, the new guard throws before `S3Url::Resolve()` can read `?s3_endpoint=...`. R2 has the same ordering problem, even though its error says the endpoint may come from the URL.
> 
> Could we handle this through a shared finalization step instead of special-casing aliases? I think the flow should be:
> 
> 1. Read secrets/settings and apply provisional provider defaults. A missing required endpoint for an alias or R2 should remain unresolved here and must not fall back to AWS.
> 2. Read URL query parameters, which have the final opportunity to supply or override the endpoint.
> 3. Call something like `FinalizeAuthParams()` to validate that aliases/R2 have a non-empty endpoint and then apply the final derived defaults.
> 
> This should preserve all of these cases:
> 
> * endpoints from a secret, `s3_endpoint`, or `?s3_endpoint=...` succeed;
> * an endpoint missing from all sources fails;
> * `?s3_endpoint=` or whitespace still fails after clearing an earlier value;
> * aliases never silently receive the AWS endpoint default.
> 
> Please add coverage that passes an alias and R2 endpoint only through the URL, plus the existing cases where a URL parameter clears an endpoint supplied by a secret or setting.
> 
> Ready to merge once this is fixed.

@lnkuiper Done as described — InitializeAuthParams now leaves a required endpoint unresolved instead of validating, url query parameters get the final say, and FinalizeAuthParams validates and applies the derived defaults; R2 goes through the same path rather than being special-cased.
Coverage added through S3Url::Resolve for alias and R2 with the endpoint supplied only by the url, plus a url parameter clearing one from a secret or setting

lnkuiper:
Thanks! LGTM now.

PR #341, opened 2026-06-16
Fall back to full GET when HEAD is not allowed

staticlibs:
Currently when `Range` requests are performed during the file download and the server does not support ranges - then the download logic falls back to use full file download with `GET`.

To perform the `Range` requests the client first sends a `HEAD` request to get the `Content-Length`. In case when `HEAD` is not supported by the server, it resorts to using `Range: bytes=0-1` requests instead of `HEAD`. The problem is that these `Range: bytes=0-1` requests (unlike normal download `Range` requests) are throwing on error, so when neither `Range` nor `HEAD` are supported - then the fallback to full `GET` does not happen.

This PR allows to fall back from `Range: bytes=0-1` responses to a full `GET` download.

Testing: adding an SQLLogic test is untrivial because `PYTHON_HTTP_SERVER_DIR` server currently used on CI only uses default `http.server` handler. A custom handler is required there to be able to disallow `HEAD` on selected requests. Please let me know if this needs to be implemented. So far the change was tested manually on OpenAI API with the following:

```sql
statement ok
CREATE TEMPORARY SECRET openai (
  TYPE http,
  SCOPE 'https://api.openai.com',
  EXTRA_HTTP_HEADERS MAP {'Authorization': 'Bearer sk-admin-xxx'}
)

query IIIIII
FROM read_json('https://api.openai.com/v1/organization/users/user-xxx')
----
user-xxx organization.user 1234567890 alexkasko@ducklabs.com Alex Kasko owner
```

Supersedes: #309

Fixes: duckdb/duckdb#21583

### DIFF
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -493,6 +493,15 @@ void HTTPFileHandle::AdaptReadBufferSize(idx_t next_read_offset) {
 	read_buffer = read_buffer.GetAllocator()->Allocate(read_buffer.GetSize() * 2);
 }
 
+static bool RespondedWithRangeRequestNotSupported(const HTTPResponse &res) {
+	if (!res.HasRequestError()) {
+		return false;
+	}
+	ErrorData error(res.GetRequestError());
+	return error.Type() == RangeRequestNotSupportedException::TYPE &&
+	       error.RawMessage() == RangeRequestNotSupportedException::MESSAGE;
+}
+
 bool HTTPFileSystem::TryRangeRequest(FileHandle &handle, string url, HTTPHeaders header_map, idx_t file_offset,
                                      char *buffer_out, idx_t buffer_out_len) {
 	auto &hfh = handle.Cast<HTTPFileHandle>();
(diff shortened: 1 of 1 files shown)


carlopi (commented):
Thanks, but I am not sure I understand what happens after setting `force_full_download`, like, how many requests are performed?

Ideally, we perform at max a single full file download, while I could see (after HEAD + GetRangeRequest) we could still do a few other GetRangeRequest then each upgraded to full downloads.

While that does improve handled situations, I am not sure if it can significantly pessimize results.

staticlibs:
Thanks for the review! 

> Thanks, but I am not sure I understand what happens after setting force_full_download, like, how many requests are performed?

If `force_download` is set by user - then the `HEAD` is not sent at all and a single download is performed.

If `force_download` is not set - then the sequence without the patch is:

 - `HEAD` -> `405`
 - `GET Range 0-1` -> no response (retried 3 times as `http_retries = 3` by default)
 - error thrown to user

With the patch:

 - `HEAD` -> `405`
 - `GET Range 0-1` -> no response (retried 3 times)
 - `GET Full` -> `200`

> Ideally, we perform at max a single full file download,

Currently this is completely manual - user needs to set `force_download`.

> while I could see (after HEAD + GetRangeRequest) we could still do a few other GetRangeRequest then each upgraded to full downloads.

M, not "each upgraded", `GET Range 0-1` is retried multiple times down the stack before throwing and triggering the fallback. The `GET Full` only happens once after 4 failed `GET Range 0-1`.

carlopi:
My question it's like:

With the patch:

HEAD -> 405
GET Range 0-1 -> no response (retried 3 times)
GET Full -> 200

But this is logically in the HEAD state. What happens when we will read the file? We will performed a separate GET on the full size or that's properly cached?

I am not sure, and double checking I think would be handy as part of the PR, if the path is:
```
HEAD
3x GET range
GET
```
or
```
HEAD
3x GET range
GET
GET
```

for cases where a logically GetFileInfo is followed by actually reading (part-of) the file.

carlopi:
Actually, the issue I was raising before doens't exist, since no GET is actually performed in the logical HEAD.

carlopi:
Thanks!

PR #343, opened 2026-06-16
Enable httpfs_connection_caching by default

npoggi:
**What:** change the default of `httpfs_connection_caching` from `false` to `true`, so HTTP(S)
requests reuse the underlying connection instead of opening a new one per request.  It tests that it is on by default and that it can be disabled via the flag.

**Why:** Without caching, every request opens a fresh TCP connection and, over HTTPS, performs a CPU-intensive TLS handshake. Reusing the connection amortizes both, improving throughput for concurrent requests by 1.3×–1.6× on plain HTTP and 1.9×–3.0× over TLS.

**Impact:** Behavior change; existing clients move to connection caching when they upgrade to the release including this change.

**Motivation:** Benchmarked using the Quack [single-row INSERT transaction benchmark](https://github.com/duckdb/duckdb-quack/blob/v1.5-variegata/benchmarks/benchmark-transactions.py) over two AWS `m8g.2xlarge` nodes in the same availability zone (AZ).

**Speedup of enabling caching:**

| client threads | plain HTTP | HTTPS (TLS) |
| -------------- | ---------- | ----------- |
| 1              | 1.6×       | 2.9×        |
| 4              | 1.4×       | 3.0×        |
| 8              | 1.3×       | 2.7×        |
| 16             | 1.3×       | 1.9×        |

image

### DIFF
--- a/src/httpfs_extension.cpp
+++ b/src/httpfs_extension.cpp
@@ -164,7 +164,7 @@ static void LoadInternal(ExtensionLoader &loader) {
 #endif
 	};
 	config.AddExtensionOption("httpfs_connection_caching", "Enable connection caching for HTTP requests",
-	                          LogicalType::BOOLEAN, Value::BOOLEAN(false), callback_httpfs_connection_caching);
+	                          LogicalType::BOOLEAN, Value::BOOLEAN(true), callback_httpfs_connection_caching);
 
 	config.AddExtensionOption("enable_global_s3_configuration",
 	                          "Automatically fetch AWS credentials from environment variables.", LogicalType::BOOLEAN,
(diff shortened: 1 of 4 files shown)


npoggi:
@carlopi could you retrigger the MinIO tests? The error looks transient: ` HTTP Error: HTTP GET error on 'https://d37ci6vzurychx.cloudfront.net/misc/taxi+_zone_lookup.csv' (HTTP 0 Internal Server Error)`.

Windows build failure seems unrelated.

carlopi:
Let's retrigger CI (expected to pass, now that duckdb commit moved forward)

carlopi:
Thanks!

PR #346, opened 2026-06-20
Fix S3 settings scope retrieval (debug only)

staticlibs:
There was a problem discovered when in assertion-enabled builds the following `ATTACH` to S3 was failing:

```sql
ATTACH 's3://my-bucket/my.duckdb'
```
```
IO Error: Cannot open database "s3://my-bucket/my.duckdb" in read-only mode: database does not exist
```

When this `ATTACH` was working correctly in release builds.

The setting retrieval assertion was triggered [here](https://github.com/duckdb/duckdb/blob/c4770ecba48065b691843da2e6eb9f91e3fea77b/src/include/duckdb/main/setting_info.hpp#L58) and was swallowed by [this catch](https://github.com/duckdb/duckdb-httpfs/blob/c3f215ab360f04dc3d3d5305fa81849c0121f111/src/httpfs.cpp#L454) afterwards, so generic IO error message was returned.

This PR adds the setting checks in two places where `GetScope` is used.

### DIFF
--- a/src/include/s3fs.hpp
+++ b/src/include/s3fs.hpp
@@ -30,8 +30,8 @@ public:
 	SettingLookupResult TryGetSecretKeyOrSetting(const string &secret_key, const string &setting_name, TYPE &result) {
 		Value temp_result;
 		auto setting_scope = reader.TryGetSecretKeyOrSetting(secret_key, setting_name, temp_result);
-		if (!temp_result.IsNull() &&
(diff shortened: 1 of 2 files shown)


carlopi (approved):
Thanks! This does make sense to me

PR should be fixed via https://github.com/duckdb/duckdb-httpfs/pull/348

carlopi:
Thanks!

PR #350, opened 2026-06-23
initialize last_modified field on HTTPFileHandle to 0

### LINKED ISSUE #23255: Malformed JSON error fetching ND-JSON twice from same URL if data changes
### What happens?

I have a script which updates data on a server and fetches an ND-JSON report before and after the update. After the update DuckDB **1.5.3** fails with an error like:
```
Malformed JSON in file "http://127.0.0.1:8000/", at byte 1 in line 4: unexpected character.
```
The line number is always two more than the number of lines in the report.
DuckDB **1.4.4** is fine.

### To Reproduce

Run this Python HTTP server:
[test_server.py](https://github.com/user-attachments/files/28886895/test_server.py)
```python
#!/usr/bin/env python3

import json
from http.server import BaseHTTPRequestHandler, HTTPServer
from io import StringIO
from itertools import count


class JsonServer(BaseHTTPRequestHandler):
    _incr = count(0)

    def do_headers(self):
        self.send_response(200)
        self.send_header("Content-Type", "application/x-ndjson")
        self.end_headers()

    def do_HEAD(self):
        self.do_headers()
        self.wfile.write(b"")

    def do_GET(self):
        self.do_headers()

        n = next(self._incr)
        rspns = StringIO()
        for i in range(2):
            rspns.write(
                json.dumps(
                    {
                        "a": i,
                        "b": "x" * n,
                    }
                )
            )
            rspns.write("\n")
        self.wfile.write(rspns.getvalue().encode("utf-8"))


if __name__ == "__main__":
    httpd = HTTPServer(("127.0.0.1", 8000), JsonServer)
    httpd.serve_forever()
```
(It changes the contents of the lines served with each response, but not the number of lines.)

and then do:
```sh
duckdb -c "FROM read_json('http://127.0.0.1:8000/'); FROM read_json('http://127.0.0.1:8000/')"
```

### OS:

MacOS

### DuckDB Version:

1.5.3

### DuckDB Client:

CLI, Python

### Hardware:

Apple M1

### Full Name:

James Gilbert

### Affiliation:

Wellcome Sanger Institute

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

- [x] Yes, I have

Yes

josephbuckingham:
Helps close https://github.com/duckdb/duckdb/issues/23255

timestamp_t is prone to initialization as garbage and would not trigger the checks implemented in https://github.com/duckdb/duckdb/pull/23427.

### DIFF
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -383,8 +383,8 @@ HTTPInput::HTTPInput(unique_ptr<HTTPParams> params_p)
 HTTPFileHandle::HTTPFileHandle(FileSystem &fs, const OpenFileInfo &file, FileOpenFlags flags,
                                shared_ptr<HTTPInput> input_p)
     : FileHandle(fs, file.path, flags), http_input(std::move(input_p)), http_params(http_input->http_params),
-      flags(flags), length(0), force_full_download(false), buffer_available(0), buffer_idx(0), file_offset(0),
-      buffer_start(0), buffer_end(0) {
+      flags(flags), length(0), last_modified(0), force_full_download(false), buffer_available(0), buffer_idx(0),
+      file_offset(0), buffer_start(0), buffer_end(0) {
 	// check if the handle has extended properties that can be set directly in the handle
 	// if we have these properties we don't need to do a head request to obtain them later
 	if (file.extended_info) {


dentiny (commented):
Thanks! Since we've missed field initialization, I'm wondering if it's better to directly assign it in the struct? The benefit of which is, it could become a coding practice so both devs and agents are easier to follow.
```cpp
struct FileHandle {
  int64_t last_mod = 0;
};
```

josephbuckingham:
That makes more sense, I'll close this.

dentiny:
> That makes more sense, I'll close this.

Ah actually I think this is important, the only thing I'd ask is whether it's better to assign default in the struct directly instead in constructor.

josephbuckingham:
Perhaps leaving it as it is and letting the classes handle it would be better. I took a quick look at the unix filesystem implementation and it handles the metadata a bit differently which might imply that it's better handled outside the struct.

dentiny (approved):

carlopi:
Thanks!

PR #352, opened 2026-06-26
Restore Content-Type emission for S3 SigV4 requests under EMSCRIPTEN

### LINKED ISSUE #351: WASM S3 PUT signs a Content-Type it never sends → 403 SignatureDoesNotMatch (regression of #105 via #234)
## Summary 

When using DuckDB-Wasm, any S3 data-file write (COPY (...) TO 's3://...', or an Iceberg INSERT/CREATE TABLE) fails with 403 SignatureDoesNotMatch on the PUT. Reads with the same credentials work. SignatureDoesNotMatch is returned at S3's auth step before any policy eval, so this is purely request-signing — not creds/IAM/CORS/region.

I am building a web app that uses duckDB wasm to attach to a Polaris catalog backed by S3 and while reads work fine, I get the following error on writes: 

Image

### Root cause 
I've already root caused this: `s3fs.cpp::CreateS3Header` signs `content-type:application/octet-stream `into the canonical request + SignedHeaders for every PUT, but the `!= "application/octet-stream"` guard suppresses emitting it. Native curl/httplib re-add the default (signed==sent); WASM's `http_wasm.cc::HTTPWasmClient::Put` sends a Uint8Array body via sync XHR with the `setRequestHeader("Content-Type", …)` lines commented out → per the XHR spec, no Content-Type on the wire. Signed application/octet-stream vs. sent nothing ⇒ 403.

I did some tracking in the issue and interestingly it was solved and then reborken — #101 added the emission (broke native) → #105 #ifdef EMSCRIPTEN-guarded it (WASM worked; @samansmink: "Won't this still break things on wasm?") → #234 replaced the guard with != "application/octet-stream", silently re-breaking WASM. No open issue tracks this.

### Reproduction 
minimal COPY (SELECT 1) TO 's3://…' on duckdb-wasm ≥ v1.30.0 → 403; DevTools shows the canonical request signing content-type:application/octet-stream with no Content-Type sent.

### Proposed fix 
emit the signed Content-Type under EMSCRIPTEN, native unchanged (the s3fs.cpp diff). Restores #105 for WASM, keeps #234's native behavior. PR to follow.

konstantinoscs:
Fixes #351.

# Summary

Restore `Content-Type` header emission for S3 SigV4 requests under `EMSCRIPTEN`.

`CreateS3Header` signs `content-type` when a content type is provided, including the `application/octet-stream` used by S3 `PUT`. After #234, that value is intentionally not emitted into the header map so native curl can keep its existing default/override behavior. In DuckDB-Wasm, however, the header map is what `HTTPWasmClient` forwards to `XMLHttpRequest`, so the browser request can be missing a header that was included in `SignedHeaders`, producing AWS S3 `403 SignatureDoesNotMatch`. Only writes are affected — `GET`/`HEAD` pass an empty content type, so it is neither signed nor sent, which is why reads work and writes fail.

This change keeps native behavior unchanged and restores the previous WASM-specific emission from #105.

# Testing
- **Native: unchanged by construction.** The `#else` branch is byte-identical to the current code, so the existing MinIO S3 suite is unaffected (regression guard).
- **DuckDB-Wasm + real S3 — validated end-to-end.** Built `httpfs` for `wasm_eh` against DuckDB `v1.5.4` (emsdk `3.1.71`) and loaded it into a DuckDB-Wasm app via `INSTALL httpfs FROM '<local-repo>'` with `SET allow_unsigned_extensions=true`. The fix lives in `s3fs.cpp`, so it applies identically to the `mvp`/`threads` variants.
  - **Before (stock httpfs):** a browser→S3-direct Iceberg `INSERT` failed with `403 SignatureDoesNotMatch` on the data-file `PUT`.
  - **After (patched httpfs):** the same `INSERT` succeeded against real S3 — the `PUT` returned `200`. In DevTools the request now carries `Content-Type: application/octet-stream`, matching the `content-type` already listed in `Authorization`'s `SignedHeaders=...` (i.e. signed == sent).
  - Isolated equivalent for reviewers (no catalog required), which exercises the same `PUT` path: `COPY (SELECT 1) TO 's3://<bucket>/<prefix>/probe.parquet' (FORMAT parquet)`.

From a web app I'm building that uses duckdb WASM to attach to a Polaris backed Catalog:
1. (Pre-fix) Reads work correctly:
image

2. (Pre-fix) Writes error (ignore the wrong error message heh)
image
with the 403 response saying:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message></Error>
```

3. (After-fix) Writes succeed 
image

4. (After-fix) Read verifies new row is there
image

# Notes

The `duckdb-httpfs` native MinIO CI is not expected to exercise this branch because the behavior is guarded by `EMSCRIPTEN`. There is also currently no CI harness that runs SQL through the WASM HTTP transport, so this path was validated manually in a browser against real S3 (above) rather than by an automated test.


konstantinoscs:
@carlopi I see that historically you are active on this area. Can you take a look?

PR #353, opened 2026-07-01
Add move constructor to CURLRequestHeaders

staticlibs:
`CURLRequestHeaders` class contains a pointer to `curl_slist` that is freed in destructor. The instance of this class is created in `TransformHeadersCurl()` and returned by value from there. Normally the copy does not happen on return due to Named Return Value Optimization. But NVRO is not guaranteed in C++17, for example it never happens in MSVC debug builds. That is causing the destructor call on return and the use-after-free for the `curl_slist` contents after that, that may or may not cause a segfault (or a hang when iterating over this list) later.

This PR adds the move constructor (and assignment operator) to `CURLRequestHeaders` and also changes the return of `TransformHeadersCurl()` to make the move constructor to be called all the time.

Ref: duckdblabs/duckdb-internal#9402

### DIFF
--- a/src/httpfs_curl_client.cpp
+++ b/src/httpfs_curl_client.cpp
@@ -495,7 +495,7 @@ private:
 				curl_headers.Add(entry.first + ": " + entry.second);
 			}
 		}
-		return curl_headers;
+		return std::move(curl_headers);
 	}
 
 	void ResetRequestInfo() {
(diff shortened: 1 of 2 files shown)


carlopi:
Thanks! LGTM

carlopi (approved):

PR #356, opened 2026-07-06
Always add user-agent to httplib request

HendrikLambert:
Closes https://github.com/duckdblabs/duckdb-internal/issues/9861

When using the (non-default) httplib implementation, the User-Agent header was missing when performing a S3 LIST operation, e.g.

```sql
SELECT count(*) FROM glob('s3://test-bucket/ua_glob_*.parquet');
```

httplib now matches the curl implementation by applying the user-agent header, when missing, on each request.

Adds httplib specific User-Agent tests and User-Agent asserts to several existing S3/HTTP tests.

### DIFF
--- a/src/httpfs_httplib_client.cpp
+++ b/src/httpfs_httplib_client.cpp
@@ -46,7 +46,14 @@ public:
 		state = http_params.state;
 	}
 
+	static void AddUserAgentIfAvailable(HTTPFSParams &http_params, HTTPHeaders &header_map) {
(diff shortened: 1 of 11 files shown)


carlopi:
Thanks! Could you move to `v1.5-variegata` branch?

HendrikLambert:
> Thanks! Could you move to `v1.5-variegata` branch?

Moved the target branch

carlopi (approved):
Thanks!

PR #358, opened 2026-07-06
Automatically try to refresh credentials on 403

lnkuiper:
And improve testing with C++ tests that include an `S3MockServer`.


dentiny:
Hey @lnkuiper , I'm wondering if you think it reasonable to have the mock server in DuckDB core repo?
- I'm debugging the CSV reader deadlock issue, but found cache only involved for [remote files](https://github.com/duckdb/duckdb/blob/c40da9ff39c905e55f1b450ae130b704bc689d0d/src/execution/operator/csv_scanner/buffer_manager/csv_file_handle.cpp#L24)
- My expectation is: we could do fanzy fuzzy testing with (1) local remote (fake) fs; (2) debug filesystem which injects deterministically random failure, latency and throttle

carlopi:
I was thinking one path to increasing testing would be having a `ChaosHTTPUtil` of sort, that wraps `CurlHTTPUtil`, but then, on getting a response, might decide at random to reject it for random reasons, and see how the system behaves.

Advantage of `ChaosHTTPUtil` is that it could live in an extension.

dentiny:
> I was thinking one path to increasing testing would be having a ChaosHTTPUtil of sort, that wraps CurlHTTPUtil

But how do we make sure it's involved in the testing? With fake filesystem extension, the filepath -> IO traffic hijack is done transparently via virtual filesystem.

> Advantage of ChaosHTTPUtil is that it could live in an extension.

Yeah that's my goal as well, the debug filesystem I mentioned above is already implemented as an in-tree extension: https://github.com/duckdb/duckdb/pull/23473

lnkuiper:
Good idea, we can do larger changes on the `main` branch, this will go into v1.5.5 so we're trying to make changes to the extensions only for this functionality

PR #363, opened 2026-07-07
Cherry-pick #335 deadlock change

dentiny:
The PR is a split for deadlock fix from https://github.com/duckdb/duckdb-httpfs/pull/335 (which contains both deadlock fix + request redirection).

### DIFF
--- a/src/http_state.cpp
+++ b/src/http_state.cpp
@@ -31,6 +31,18 @@ void CachedFileHandle::AllocateBuffer(idx_t size) {
 	file->capacity = size;
 }
 
+void CachedFileHandle::ResetBuffer() {
+	if (file->initialized) {
+		throw InternalException("Cannot reset a buffer for a cached file that was already initialized");
+	}
+	if (!lock) {
+		throw InternalException("Cannot reset a buffer for a cached file without lock");
+	}
+	file->data.reset();
+	file->capacity = 0;
+	file->size = 0;
+}
+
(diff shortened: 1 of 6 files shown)


carlopi:
I was reviewing this: should this go into `v1.5-variegata` as well?

lnkuiper:
I think the caching file system is very different on v1.5, so my guess is no but @dentiny knows best

carlopi:
I am not sure, I got this assessment

> It should go in: I proved the deadlock is still reachable on v1.5 by forcing a second region redirect, which slips past the one-shot retried_region guard in v1.5's request-layer handler and lands in the unguarded retry at the Initialize level. The unfixed
  build hangs on std::mutex::lock inside CachedFileHandle's constructor; the cherry-picked build returns its rows and exits zero.

But indeed, let's check with @dentiny.

dentiny:
Hey! I think both https://github.com/duckdb/duckdb-httpfs/pull/364 and this PR https://github.com/duckdb/duckdb-httpfs/pull/363 should be contained in v1.5 release, each of them fixing one issue.
The region redirect logic might be needed as well: https://github.com/duckdb/duckdb-httpfs/pull/335 
From my understanding, we need to point [DuckDB core commit](https://github.com/duckdb/duckdb/blob/86eec92365100073dcc69c50f8d5ae14e65e824f/.github/config/extensions/httpfs.cmake#L4) to the latest httpfs extension commit before release, right?

PR #365, opened 2026-07-10
Retry transient S3 RequestTimeout (400) on multipart part upload

guillesd:
S3 returns HTTP 400 with <Code>RequestTimeout</Code> (not 408) when a multipart part's socket stalls past its idle timeout under upload contention. Core's retry classifier (HTTPResponse::ShouldRetry) is status-code-only and covers 408/429/5xx, so this transient failure surfaces to users as a fatal Bad Request (HTTP code 400) and http_retries never engages.

Since the retry loop lives in duckdb core and the discriminator (RequestTimeout) only exists in the S3 XML body, this fixes it in the S3 layer: a part PUT is idempotent, so UploadBufferImplementation retries this specific 400 using the existing http_retries / http_retry_wait_ms / http_retry_backoff settings. Generic 400s still fail fast. Basically everything else goes through core's machinery for retrying, even other 400s.

**Testing:** Adds fault injection to the mock S3 server and cpp unit tests for both the retried RequestTimeout 400 and a non-retryable generic 400.

### DIFF
--- a/src/s3_multi_part_upload.cpp
+++ b/src/s3_multi_part_upload.cpp
@@ -131,8 +131,7 @@ void S3MultiPartUpload::UploadBufferImplementation(shared_ptr<S3WriteBuffer> wri
 	string etag;
 
 	try {
-		// Transient S3 errors (e.g. RequestTimeout) are retried inside PutRequest by the shared S3 request
-		// loop; a non-OK status here is therefore already final.
+		// Transient S3 errors are retried inside PutRequest; a non-OK status here is already final.
 		res = s3fs.PutRequest(*http_input, path, {}, (char *)write_buffer->Ptr(), write_buffer->idx, query_param);
 
 		if (res->status != HTTPStatusCode::OK_200) {
--- a/src/s3_multi_part_upload.cpp
+++ b/src/s3_multi_part_upload.cpp
@@ -1,5 +1,7 @@
 #include "s3_multi_part_upload.hpp"
 
+#include "duckdb/common/chrono.hpp"
+
 #include <thread>
 #ifdef EMSCRIPTEN
 #define SAME_THREAD_UPLOAD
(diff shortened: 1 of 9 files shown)


src/s3_multi_part_upload.cpp
```diff
@@ -125,17 +127,39 @@ void S3MultiPartUpload::UploadSingleBuffer(shared_ptr<S3WriteBuffer> write_buffe
 	UploadBufferImplementation(write_buffer, "", true);
 }
 
+// S3 reports a stalled part upload as 400 RequestTimeout (not 408), so core's status-based retry
+// misses it. Re-PUTting a part is idempotent, so retry this specific 400 locally.
+static bool IsTransientS3UploadError(const HTTPResponse &res) {
```
dentiny: BTW I don't think it's an upload-specific error code, seem to be general
https://docs.aws.amazon.com/AmazonS3/latest/developerguide/ErrorResponses.html#ErrorCodeList

Error description
```
Your socket connection to the server was not read from or written to within the timeout period.
```

dentiny: OpenDAL is the most battle-tested storage backend accessor I've seen, here's the error categorization, just for reference :)
https://github.com/apache/opendal/blob/ae99a3b016e354a1b2bb2baf0c70f9f9e134970a/core/services/s3/src/core.rs#L2044-L2057

guillesd: I see your point, this is generic for all S3. So we should potentially go for a better more generic fix... let me discuss this internally

guillesd: Looked at it and added it in some abstraction @lnkuiper built to handle auth of all s3 requests.

lnkuiper: Discussed internally to improve the generic fix

guillesd:
@lnkuiper should be good now (hopefully)

carlopi (changes requested):
I took a look, I think I found these possible improvement points:
 
#### 1. [correctness] `FindTagContents` can throw `InternalException` while classifying a 400 body
`src/s3fs.cpp:422` (`ParseTransientS3Error` → `FindTagContents(body,"Code",…)`)`FindTagContents` (s3fs.cpp:1821-1824) throws `InternalException` when it finds an opening `<Code>` with no matching `</Code>`. Failure scenario: an S3-compatible backend / proxy / truncated response returns a 400 (or 404/500…) whose body contains `<Code>` but no closing tag. Before this PR that response threw a normal `HTTPException`; now `FindTagContents` throws `InternalException`, which **invalidates the database** (subsequent queries fail), and — on the throw path — replaces the original HTTP error with the internal one.
Reachability is low (real AWS/minio always send balanced XML; the realistic trigger is a mid-body connection cut), but it is a strictly-worse failure mode than the status quo.

Same mechanism at `src/s3_multi_part_upload.cpp:139`: the new `S3FileSystem::ParseS3Error(res->body)` in the throw-expression can also throw `InternalException` on a malformed body (note: `ParseS3Error` is already called this way elsewhere via `GetS3Error`, so this call site is not novel — but it shares the exposure).

Direction: make `FindTagContents` (or a wrapper used for classification) tolerant of an unmatched tag — return "not found" instead of throwing — at least on the transient-classification path, since a malformed error body should never escalate to a DB-invalidating internal error.

#### 2. [portability] `SleepForTransientRetry` calls `ThreadUtil::SleepMs` unguarded `src/s3fs.cpp:451-459`

Core's own retry sleep is wrapped in `#ifndef DUCKDB_NO_THREADS` (duckdb `http_util.cpp`
`RunRequestWithRetry`). The PR's `SleepForTransientRetry` is not. On a `DUCKDB_NO_THREADS` build (e.g. the
wasm MVP target), `ThreadUtil::SleepMs` **throws `InvalidInputException`** ("requires DuckDB to be compiled
with thread support"). The first retry (`transient_retries == 0`) doesn't sleep, so it's fine; the **second+
retry** hits `SleepMs` and throws.

Direction: mirror core — guard the sleep with `#ifndef DUCKDB_NO_THREADS`, or skip sleeping when threads
are unavailable.

#### 3. [efficiency/correctness] Transient retry reuses the (possibly poisoned) cached HTTP client `src/s3fs.cpp:478-511` vs `SendS3HandleRequestWithClientCache` (s3fs.cpp:561-569)

Core's `RunRequestWithRetry` re-initializes the HTTP client before each retry (its `on_retry` cb). The PR's transient loop just `continue`s; for the handle-request path the request goes back through `SendS3HandleRequestWithClientCache`, which pulls a client from `client_cache` — potentially the same keep-alive connection that just stalled. Since `RequestTimeout` *is* a stalled-socket condition, the retry may immediately re-fail on the same dead connection, making the fix less effective than intended for its own target scenario. Not a bug — the retry is still bounded and correct — but worth confirming whether the reused connection is actually refreshed. (I did not manage to prove the connection is stale on retry; flagging as a question, not a confirmed defect.)


I think 2 must just be fixed, 3 I think needs re-initialization or anyhow having a look at that, 1 seems more theoretical than anything, but could also be fixed.

guillesd:
@carlopi should be okey now. Thanks for the catch(es)

carlopi:
I thought about this quite a bit more, I realized there is now another issue: 2 rounds of retries, potentially (as long as code drift) nesting each other. And the other one is complexity, this is implementing a second retry loop (for all requests types, not only PUT) that adds complexity that might be better folded like:
* land `duckdb/duckdb`'s https://github.com/duckdb/duckdb/pull/23793 (let's check what maintainers thinks)
* slim this PR down with regard to the retry loop / retry machinery, and instead overriding then new hook `ShouldRetry` to detect 400 + timed out.

Also a note for later myself, `httplib` had a different behaviour since all is currently retried via try-catch. This is on the weird side, and I hadn't realized until now, so some though might be required in handle `httplib`.

I also realized, but I will land this in a side, that there is an issue with exhaustion of physical connect resources, that might be better land as a separate PR.

lnkuiper:
I've merged https://github.com/duckdb/duckdb/pull/23793 this should update the submodule and rebase the fix on that infrastructure

guillesd:
Hi @lnkuiper @carlopi I've updated the submodule and the code to reuse the new stuff from httputil. Can you check? Do I need to update anything in the `MainDistributionPipeline` with the submodule bump? I know we do this differently in some extensions

lnkuiper:
Thanks!

PR #367, opened 2026-07-16
Bump submodules to the latest DuckDB

staticlibs:
This PR updates the DuckDB submodule to the latest `main` and applies the following patches:
    
 - `0007-logging-storage-includes.patch`
 - `0008-logger-secret-extension.patch`
 - `0009-global-connection-pool.patch`
    
Other patches (`0001` - `0006` and `0010`) appeared to be already applied.
    
The intention is to follow-up with the `httpfs` update in the `duckdb/duckdb`.
    
Ref: duckdblabs/duckdb-internal#5731

### DIFF
--- a/src/create_secret_functions.cpp
+++ b/src/create_secret_functions.cpp
@@ -1,4 +1,5 @@
 #include "create_secret_functions.hpp"
+#include "duckdb/logging/logger.hpp"
 #include "s3fs.hpp"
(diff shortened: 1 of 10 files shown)


lnkuiper:
Thanks! I've done a lot of patching and changing over the last few weeks, this is no longer needed

PR #368, opened 2026-07-17
Merge v1.5 into main

Tishj:
Opening this as it's a blocker for upstreaming duckdb-iceberg v1.5 changes into main

There were quite a couple merge conflicts, I let jesus (AI) take the wheel, with the aim to let the tests tell us whether something fell through the cracks, at which point I'll investigate.


Tishj:
[This](https://github.com/duckdb/duckdb-httpfs/pull/197) is causing tests to fail, because the tests have side effects that aren't cleaned up, but we rerun them

Tishj:
The integration tests failed but the workflow swallowed the error, working on a fix

lnkuiper:
I think the merge is clean, but Codex found some pre-existing issues on `v1.5-variegata`. I'll try to send a PR for those

PR #372, opened 2026-07-21
Fix #370 and include fix from #354

### LINKED ISSUE #370: curl backend: range-GET timeout + retry can silently return empty values instead of erroring (deterministic repro)
While investigating remote-parquet read failures in an ETL pipeline (DuckDB streaming parquet from object storage into a slower sink), we found a case where the curl httpfs backend returns **wrong query results with no error**. Everything below is measured on DuckDB v1.5.3 (`14eca11bd9`, osx_arm64, httpfs from current main); the repro is fully local, no credentials required.

### Repro (3 steps)

**1.** Generate a 1 GB parquet whose column chunks are large (~205 MB — wide VARCHAR, uncompressed, unique values so nothing dedupes):

```sql
COPY (SELECT i, repeat('x', 99990) || lpad(i::VARCHAR, 10, '0') AS payload
      FROM range(10000) t(i))
TO 'big.parquet' (ROW_GROUP_SIZE 2000, COMPRESSION 'UNCOMPRESSED');
```

**2.** Serve it over HTTP rate-limited to 6 MB/s per request (simulates constrained per-stream bandwidth; any Range-supporting throttled server works — script below):

```bash
python3 slow_server.py big.parquet 8766 6
```

At 6 MB/s a 205 MB chunk needs ~34 s — just over the default `http_timeout` of 30 s.

**3.** Read it:

```sql
SELECT COUNT(*), SUM(strlen(payload))
FROM read_parquet('http://127.0.0.1:8766/big.parquet');
-- expected: 10000, 1000000000
-- actual:   10000,  936000000     (3/3 runs, identical, ~121 s, NO error)
```

`WHERE strlen(payload) <> 100000` shows the damage is exactly the **last 160 rows of each of the four full-size ~205 MB column chunks** (640 rows total, e.g. 1888–2047), all read back as **empty strings** — i.e. the timed-out GET delivered all but the tail of each chunk and the missing bytes were materialized as empty values rather than an error.

### Same workload on other configurations (all measured, sequential, same host)

| config | result |
|---|---|
| 1.5.3, curl (default) | `936000000`, no error, 121 s — **silent corruption** |
| 1.5.3, `SET httpfs_client_implementation='httplib'` | correct, 32 s |
| 1.4.4 (defaults) | correct, 33 s |
| 1.5.3, curl, `SET http_timeout=600000` | correct, 33 s |

### Why the boundary sits at 1.5, as far as we can tell

The curl client passes `http_timeout` to `CURLOPT_TIMEOUT` (`httpfs_curl_client.cpp:186`), libcurl's *total-operation* cap, while the httplib client passes it to `set_read_timeout` (`httpfs_httplib_client.cpp:27`), an *idle* timeout. When 1.5 made curl the default, any request whose total duration exceeds 30 s started failing — and on this path, apparently not always loudly. We separately reproduced the loud flavor (`Timeout was reached error for HTTP GET`, full query abort) on cloud-hosted files — that part looks like #334 — but the silent flavor above seems strictly worse, hence its own issue.

Two observations that may help place the fix:

- the curl client already avoids wall-clock caps for uploads (`CURLOPT_TIMEOUT=0` + `CURLOPT_LOW_SPEED_TIME/LIMIT`, `httpfs_curl_client.cpp:325-327`, "no hard timeout for uploads") — GETs never got the same treatment;
- retried ranges restart from byte 0 (we saw the same ~250 MB range fetched 3× in an `enable_logging('HTTP')` trace on a related workload), so under sustained low bandwidth retries burn transfer without progressing. Resuming from the last received offset would make retries progress-making; independent of that, a short read should surface as an error, never as empty values.

Happy to open a PR for the GET timeout semantics and/or the short-read check.

<details><summary>slow_server.py (rate-limited HTTP server with Range support)</summary>

```python
"""Minimal HTTP server with Range support, rate-limited to simulate
constrained per-stream bandwidth. Usage: python3 slow_server.py <file> <port> <MB/s>"""
import http.server
import os
import re
import sys
import time

FILE, PORT, RATE = sys.argv[1], int(sys.argv[2]), float(sys.argv[3]) * 1024 * 1024
SIZE = os.path.getsize(FILE)
CHUNK = 256 * 1024


class Handler(http.server.BaseHTTPRequestHandler):
    protocol_version = "HTTP/1.1"

    def log_message(self, *a):
        pass

    def do_HEAD(self):
        self.send_response(200)
        self.send_header("Content-Length", str(SIZE))
        self.send_header("Accept-Ranges", "bytes")
        self.end_headers()

    def do_GET(self):
        start, end = 0, SIZE - 1
        m = re.match(r"bytes=(\d+)-(\d*)", self.headers.get("Range", ""))
        if m:
            start = int(m.group(1))
            if m.group(2):
                end = int(m.group(2))
            self.send_response(206)
            self.send_header("Content-Range", f"bytes {start}-{end}/{SIZE}")
        else:
            self.send_response(200)
        length = end - start + 1
        self.send_header("Content-Length", str(length))
        self.send_header("Accept-Ranges", "bytes")
        self.end_headers()
        t0 = time.time()
        sent = 0
        with open(FILE, "rb") as f:
            f.seek(start)
            while sent < length:
                data = f.read(min(CHUNK, length - sent))
                if not data:
                    break
                try:
                    self.wfile.write(data)
                except BrokenPipeError:
                    return
                sent += len(data)
                ahead = t0 + sent / RATE - time.time()
                if ahead > 0:
                    time.sleep(ahead)


http.server.ThreadingHTTPServer(("127.0.0.1", PORT), Handler).serve_forever()
```

</details>

lnkuiper:
Supersedes #354

Tested with `MockS3Server`

### DIFF
--- a/src/http_state.cpp
+++ b/src/http_state.cpp
@@ -98,13 +98,18 @@ void HTTPState::WriteProfilingInformation(std::ostream &ss) {
 }
 
 //! Get cache entry, create if not exists
-shared_ptr<CachedFile> &HTTPState::GetCachedFile(const string &path) {
+shared_ptr<CachedFile> HTTPState::GetCachedFile(const string &path) {
 	lock_guard<mutex> lock(cached_files_mutex);
 	auto &cache_entry_ref = cached_files[path];
 	if (!cache_entry_ref) {
 		cache_entry_ref = make_shared_ptr<CachedFile>();
 	}
 	return cache_entry_ref;
+}
+
+void HTTPState::EraseCachedFile(const string &path) {
+	lock_guard<mutex> lock(cached_files_mutex);
+	cached_files.erase(path);
 }
 
 } // namespace duckdb
--- a/src/httpfs.cpp
+++ b/src/httpfs.cpp
@@ -357,7 +357,14 @@ unique_ptr<HTTPResponse> HTTPFileSystem::RunGetRangeRequest(HTTPFileHandle &hfh,
 	    });
 
 	get_request.try_request = auto_fallback_to_full_file_download;
-	return send_request(get_request);
+	auto response = send_request(get_request);
+	if (response && !response->HasRequestError() && response->Success() && buffer_out != nullptr &&
+	    out_offset != buffer_out_len) {
+		throw IOException("Short read for HTTP GET to '%s': requested range %s (%llu bytes), but received %llu bytes",
+		                  url, range_expr, static_cast<unsigned long long>(buffer_out_len),
+		                  static_cast<unsigned long long>(out_offset));
+	}
+	return response;
 }
 
 unique_ptr<HTTPResponse> HTTPFileSystem::PostRequest(HTTPInput &input, string url, HTTPHeaders header_map,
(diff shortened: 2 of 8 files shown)


carlopi (approved):
LGTM, thanks!

Issue #334, opened 2026-05-26
Long-running parallel parquet reads from public S3 fail on 1.5.x (work on 1.4.4)


lixun910:
### What happens

On `duckdb==1.5.3` (`osx_arm64`, `httpfs 52afb42`), running a long-running
parquet read against AWS S3 fails partway through with one of the following
errors:

```
IO Error: SSL connect error error for HTTP GET to 'https://<bucket>.s3.<region>.amazonaws.com/.../*.parquet'
IO Error: SSL connection failed error for HTTP GET to 'https://<bucket>.s3.<region>.amazonaws.com/.../*.parquet'
IO Error: Timeout was reached error for HTTP GET to 'https://<bucket>.s3.<region>.amazonaws.com/.../*.parquet'
HTTP Error: HTTP GET error on 'https://<bucket>.s3.<region>.amazonaws.com/.../*.parquet' (HTTP 0 Internal Server Error)
```

The exact wording varies run-to-run on the same query. The first dozens of
parquet `Range`-GETs return 206 Partial Content fine; failures only occur
once the parquet reader has been streaming many parallel range GETs for a
minute or two.

The same query against the same data succeeds on `duckdb==1.4.4`.

We originally hit this on an authenticated workload against an Iceberg REST
catalog whose data files live in S3, but it reproduces just as cleanly on a
public bucket with no credentials, so credentials/Iceberg are not part of
the problem.

### Versions where reproduced

| duckdb | httpfs                                                   | result                              |
| ------ | -------------------------------------------------------- | ----------------------------------- |
| 1.4.4  | (default `httplib`)                                      | ✅ succeeds, ~9 min full read       |
| 1.5.3  | 52afb42 (default `curl`)                                 | ❌ fails @ ~80–160 s (varies)       |
| 1.5.3  | 52afb42 + `SET httpfs_client_implementation = 'httplib'` | ❌ fails @ ~150 s (delayed, varies) |

So the issue is **not** specific to the new curl-based HTTP backend introduced
in [1.5.0](https://duckdb.org/2026/03/09/announcing-duckdb-150.html#network-stack):
both backends fail in 1.5.3, just at different latencies. `httplib` only delays
the failure threshold; for queries that take a few minutes it still fails.

Network is fine: `curl -v https://overturemaps-us-west-2.s3.us-west-2.amazonaws.com/`
from the same host completes a TLS handshake cleanly with the system CA store,
[...]

PR #377, opened 2026-07-27
Fall back to path-style URLs for dotted S3 bucket names

DinosL:
A bucket whose name contains dots is not addressable as a vhost over TLS and [AWS documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html) acknowledges this. This PR adds a fall-back to path style for dotted S3 buckets.

This issue was reported here: https://github.com/duckdb/duckdb-httpfs/issues/334#issuecomment-5057505311. It is only reproducible with the `curl` client, which verifies certificates by default. On the contrary, `httplib`'s default value for certificate verification is false. Not sure why we have two different settings (`enable_server_cert_verification` and `enable_curl_server_cert_verification`) with different default values. I will look into it and potentially unify/deprecate one of them, but that's for another PR.


lnkuiper:
Thanks!

zaneselvans:
Thank you!

PR #388, opened 2026-08-03
Add extra_http_headers setting to inject headers into every request

HendrikLambert:
Headers can currently only be attached via a secret, and only one secret applies per request, so headers on an HTTP secret never reach traffic matching an S3 secret. This adds an `extra_http_headers` setting (`MAP(VARCHAR, VARCHAR)`) that supplies the base headers for every request, a matching secret still overrides them per key. 
One use case is testing: a suite routing traffic through a shared proxy can tag every request with a per-test header, regardless of which secret each test happens to use.

NOTE: `extra_http_headers` does not yet autoload `httpfs`

### DIFF
--- a/src/http/http_settings.cpp
+++ b/src/http/http_settings.cpp
@@ -25,6 +25,19 @@ static void SetCACertFile(ClientContext &context, SetScope, Value &parameter) {
 	parameter = Value(fs.CanonicalizePath(value, opener));
 }
 
+static void SetExtraHTTPHeaders(ClientContext &, SetScope, Value &parameter) {
+	if (parameter.IsNull()) {
+		throw InvalidInputException("NULL is not a valid option for extra_http_headers");
+	}
+	for (const auto &entry : MapValue::GetChildren(parameter)) {
+		auto key_value = StructValue::GetChildren(entry);
+		if (key_value[1].IsNull()) {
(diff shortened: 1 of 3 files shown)


src/http/http_settings.cpp
```diff
@@ -73,6 +86,10 @@ void HTTPSettings::Register(DBConfig &config) {
 	    "http_keep_alive",
 	    "Keep alive connections. Setting this to false can help when running into connection failures",
 	    LogicalType::BOOLEAN, Value(HTTPParams::DEFAULT_KEEP_ALIVE));
+	config.AddExtensionOption("extra_http_headers", "Extra HTTP headers added to every request",
```
lnkuiper, line 89: Just a tiny nitpick, can we rename this to `http_extra_headers` so that it's prefixed with `http_` like some of the other settings?

lnkuiper (commented):
Thanks! One comment:

src/http/http_settings.cpp
HendrikLambert: Thank you for the review. One small note on that is that a `SECRET` also has an `extra_http_headers` option, which functions the same as this setting, only scoped differently. `http_extra_headers` would be more descriptive as a setting, but it would then diverge from the secret option naming. Let me know which you prefer

lnkuiper: OK, let's keep it like this then

PR #399, opened 2026-08-12
Cache certificate store per connection

smvv:
### Context

I used samply to profile this query:
```sql
SELECT count(*)
FROM 'https://blobs.duckdb.org/data/taxi_2019_04.parquet'
WHERE pickup_at BETWEEN '2019-04-15' AND '2019-04-20';
```
and noticed CPU spikes in multiple threads:

image

### Certificates are parsed for each new connection

`X509_load_cert_crl_file_ex` is called for each new TLS connection, and the CA store is re-use when a connection is used a second time.

However, we can also load the CA store once and re-use it for all (new) TLS connections. This lowers the CPU cost and connection setup latency for the remaining TLS connections.

### Caching 

The shared `CurlCertificateStoreCache` can be re-used across cURL clients and handles. Synchronizes cache access so concurrent requests load each certificate store only once. Reference-counts stores with `X509_STORE_up_ref`, allowing multiple SSL contexts to use them safely
 
It caches OpenSSL `X509_STORE` objects by file path, and uses file size, modification time to file changes. Reloads a cached store when the file changes or after a 24-hour timeout.

### When is CA store not cached?
- Fall back to `CURLOPT_CAINFO` if the OpenSSL version is not compatible.
- Skips CA store loading when SSL verification is disabled.

### Unrelated commits

This PR builds on top of https://github.com/duckdb/duckdb-httpfs/pull/331. Ignore those commits.


carlopi:
Unsure on "
Use CURLSSLOPT_NATIVE_CA only if we dont enable the certificate store", maybe then there needs to be a way to opt-out of this, since `CURLSSLOPT_NATIVE_CA` would be handy to have.

carlopi:
Maybe a simpler patch: what if we just conditionally remove `NATIVE_CA` (if not opted-in explicitly), that should give back curl-handled caching, for way less complexity.

lnkuiper:
> Maybe a simpler patch: what if we just conditionally remove NATIVE_CA (if not opted-in explicitly), that should give back curl-handled caching, for way less complexity.

curl’s CA cache isn’t shared across separate easy handles, so I think this is needed for parallel reads

src/http/httpfs_curl_client.cpp
```diff
@@ -92,16 +96,26 @@ CURLHandle::CURLHandle() {
...
+			SetOption(CURLOPT_SSL_CTX_FUNCTION, ConfigureSSLContext);
+			SetOption(CURLOPT_SSL_CTX_DATA, this);
+		}
 	}
 	long ssl_options = CURLSSLOPT_AUTO_CLIENT_CERT;
-	if (use_native_ca) {
+	if (use_native_ca && !certificate_store_cache) {
 		ssl_options |= CURLSSLOPT_NATIVE_CA;
 	}
 	SetOption(CURLOPT_SSL_OPTIONS, ssl_options);
```
carlopi: I think this block is still off, there needs to be a way, user side, to say they want to use `CURLSSLOPT_NATIVE_CA` and that to be propagated, since this is basically disabling that feature but somewhat hidden.

lnkuiper:
Alternatively, we can wait until we do multi curl. Those requests share certificates by default

carlopi (commented):
I still think this is quite some heavy machinery, and on a critical path for security. Code is at the moment solid, but failure mode is bad, and I would prefer not to go down this path.

I would prefer the alternative fix is better here:
1. have an option guarding `CURLSSLOPT_NATIVE_CA` user side, by default it should not be used
2. returning (via parking lot) pre-initialized single handles to be reused (this should be there already)
3. when we get to multi-curl, that's already solved by using multi handles

lnkuiper:
1. Prefer not to add another setting
2. Indeed it's fixed after the first wave of requests (but would be nice for, e.g., lambda's to not have this cost)
3. Long-term multi curl is the cleanest solution, that will be for v2.1.0

carlopi:
I disagree on "Prefer not to add another setting", in the sense that there is a tradeoff, where everyone is paying for a feature they will not use, so this has to become opt-in in my view. I can try to think if there is a combination of current settings to express this cleanly, but basically turning on `|= CURLSSLOPT_NATIVE_CA` disables certificates sharing for the same handle.

I would just run the same experiment as above without the `|= CURLSSLOPT_NATIVE_CA` line, if most of of the cost disappears, then that's is to me the solution.

Also note that the PR as is does disable CURLSSLOPT_NATIVE_CA in the main path, and that's the thing that enables the gains.

lnkuiper:
The setting will become obsolete in the next release if we start using multi-curl, and most people won't bother to opt-in (DuckDB's default settings should be good). We're trying to remove settings, not add new ones. I'm OK with improving the default behaviour, or just waiting until v2.1.

Issue #380, opened 2026-07-30
`CreateS3SecretFunctions::GenerateRefreshSecretInfo` doesn't support `TRANSACTION` persist_type


Tishj:
Currently, the function will always set `result.persist_type = SecretPersistType::TEMPORARY;`

This is causing a refresh for a [TRANSACTION](https://github.com/duckdb/duckdb/pull/23882) scoped secret to fail: https://github.com/duckdb/duckdb-iceberg/actions/runs/30473069693/job/90649416688?pr=1186

> SELECT * FROM vended_refresh.default.vended_refresh_table ORDER BY ALL;
================================================================================
Actual result:
================================================================================
Invalid Input Error: Exception thrown while trying to refresh secret "da16e264-c1ec-456f-8420-c255a716ad09__0". To fix this, please recreate or remove the secret and try again. Error: 'Invalid Input Error: Transaction secrets do not support explicit storage'

Or I guess more specifically:
> result.storage_type = secret_entry.storage_mode;

Backtrace:
```
CreateS3SecretFunctions::TryRefreshS3Secret
```
reaches:
```
GenerateRefreshSecretInfo
```

This sets:
```c++
	CreateSecretInput result;
	result.on_conflict = OnCreateConflict::REPLACE_ON_CONFLICT;
	result.persist_type = SecretPersistType::TEMPORARY;

	result.type = kv_secret.GetType();
	result.name = kv_secret.GetName();
	result.provider = kv_secret.GetProvider();
	result.storage_type = secret_entry.storage_mode;
	result.scope = kv_secret.GetScope();
```
and returns, reaching next:
```c++
auto res = secret_manager.CreateSecret(context, refresh_input);
```
Which finally gets here:
```c++
	// Call the function
	auto secret = function_lookup->function(context, function_input); // <--- this is where we get to create our secret

	if (!secret) {
		throw InternalException("CreateSecretFunction for type: '%s' and provider: '%s' did not return a secret!",
		                        input.type, input.provider);
	}

	// Register the secret at the secret_manager
	return RegisterSecretInternal(transaction, std::move(secret), input.on_conflict, input.persist_type,
	                              input.storage_type); // <-- this is where the secret manager overrides our settings with whatever was constructed by `GenerateRefreshSecretInfo`
```

```c++
typedef unique_ptr<BaseSecret> (*create_secret_function_t)(ClientContext &context, CreateSecretInput &input);
```
[...]

PR #413, opened 2026-08-30
Refresh extension-storage S3 secrets without writing to that storage

### LINKED ISSUE #195: Using STS assume role with persistent secret fails
I have the following persistent secret configured on session initialization. 

```
CREATE OR REPLACE PERSISTENT SECRET persistent_credential_chain (
      TYPE s3,
      PROVIDER credential_chain,
      CHAIN 'sts',
      assume_role_arn 'arn:aws:iam::x:role/my-role',
      external_id 'generic-role',
      region 'us-west-2'
  );
```

This configuration results in the following error. 
```
Exception thrown while trying to refresh secret persistent_credential_chain. To fix this, please recreate or remove the secret and try again. Error: 'Invalid Input Error: Cannot create temporary secrets in a persistent secret storage!'
```

This looks like a bug to me. When using `sts`, credentials are automatically refreshed and it seems that it replace existing credentials with persistent storage.

### LINKED ISSUE #412: Secret refresh re-creates the secret in its origin storage, which breaks (or silently mutates) extension-registered secret storages
### What happens

`TryRefreshS3Secret` refreshes a secret by generating a `CreateSecretInput` from the existing entry and calling `SecretManager::CreateSecret`. `GenerateRefreshSecretInfo` copies the entry's persistence and storage:

```cpp
result.persist_type = secret_entry.persist_type;
...
if (result.persist_type != SecretPersistType::TRANSACTION) {
    result.storage_type = Identifier(secret_entry.storage_mode);
}
```

(`src/create_secret_functions.cpp`, `GenerateRefreshSecretInfo`)

For secrets in the built-in `memory`/`local_file` storages this is fine. For secrets loaded from an **extension-registered storage** — e.g. a PostgreSQL-backed store attached with duckdb-postgres's `SECRET_STORAGE_TABLE` — re-creating into the origin storage is wrong in both possible directions:

1. **Read-only storage session** → the refresh throws (surfaced as `Exception thrown while trying to refresh secret <name> …`, or `Cannot create temporary secrets in a persistent secret storage` depending on persist-type mismatch), so a read that could have succeeded after re-resolving credentials fails instead.
2. **Writable storage session** → the refresh **silently writes the secret back to the shared store**. A session-local credential re-resolution becomes a durable, multi-user mutation. We have observed a long-lived session *resurrecting rows that had been deleted from the store* hours earlier, because its loaded copy refreshed and re-persisted them.

Refreshed credential material is inherently session-local state (the point of `credential_chain` + `refresh` is "my resolved credentials went stale — resolve again"); persisting it to a shared backing store is never the right side effect, and failing the read because the store is read-only is also not the right outcome.

### Reproduction sketch

Deterministic without waiting for STS expiry — a wrong-key secret with `refresh 'auto'` triggers the refresh path on the first 403:

```sql
-- store attach per duckdb-postgres:
ATTACH '<conninfo>' AS store (TYPE postgres, SECRET meta,
       SECRET_STORAGE_TABLE 'duckdb.duckdb_secrets');
-- a secret that lives in that storage, with refresh enabled
--   (created via a writable session, or seeded externally)
-- then, in a session whose store role is read-only:
FROM 's3://some-bucket/some.parquet';
-- → auth failure → TryRefreshS3Secret → CreateSecret(storage_type='postgres_store')
--   → refresh exception instead of a re-resolve
-- in a writable session: the row's contents/timestamp change in the store
```

### Suggested fix

Refresh should update the loaded secret's credential material **in place**, without any storage write: a `SecretManager` API that swaps the cached entry's contents transiently, which `TryRefreshS3Secret` calls instead of `CreateSecret`. The existing `// TODO: change SecretManager API to avoid requiring catching this exception` in `TryRefreshS3Secret` points at the same seam, and an in-place update would remove that exception-catching along the way.

Re-creating the refreshed secret into `memory` storage instead would avoid the storage write but leave the same name live in two storages at once — the by-name ambiguity state that `GetSecretByName` and `DROP SECRET` treat as an error — so in-place is the direction that fixes refresh without introducing a new inconsistency.

### Related

- #195 reports the read-only half of this on the built-in persistent storage (`Cannot create temporary secrets in a persistent secret storage` for an `sts`-chain persistent secret) — same root cause, seen from the storage that can only fail loudly. An in-place refresh would resolve it too.
- The current copy of `persist_type`/`storage_mode` into the refresh's `CreateSecretInput` appears to originate in the fix for #380 (previously the refresh hardcoded `TEMPORARY`). That fix was right for TRANSACTION-scoped secrets, but for writable non-built-in storages it makes the refresh's `CreateSecret` land as a durable write — the silent-mutation half above.

kinghuang:
Fixes #412.

## Problem

`TryRefreshS3Secret` re-creates the refreshed secret into its origin storage (`result.storage_type = Identifier(secret_entry.storage_mode)`). For a secret loaded from an **extension-registered storage** — e.g. a database-backed store such as duckdb-postgres' `SECRET_STORAGE_TABLE` — that turns a session-local credential re-resolution into a shared-store write:

* **read-only store session** → the re-create throws (`Exception thrown while trying to refresh secret …`), so a read the fresh credentials could have served fails instead;
* **writable store session** → the refresh silently persists to the shared store. We observed a long-lived session resurrecting rows another session had deleted hours earlier.

## Fix

Route the refresh to the transaction-scoped storage when the origin storage is not one built into the secret manager:

```cpp
if (result.persist_type == SecretPersistType::TRANSACTION || !RefreshMayWriteToStorage(secret_entry.storage_mode)) {
	result.persist_type = SecretPersistType::TRANSACTION;
} else {
	result.storage_type = Identifier(secret_entry.storage_mode);
}
```

`TransactionSecretStorage` fits within today's core API: it performs no `StoreSecret` on any real storage; its tie-break offset is 0 and it is consulted first, so the refreshed copy wins the lookup `ReloadS3AuthMaterial` then performs; and `GetSecretByName`/`DropSecretByName` short-circuit on it, so two live copies of one name raise no ambiguity error — which a `memory`-storage copy would. The existing `TRANSACTION` branch from #380 is subsumed by the same condition.

`memory`, `connection`, `transaction` and `local_file` are unchanged, so no existing test changes.

## Two notes for reviewers

**Should `local_file` stop persisting too?** The same argument applies, more weakly: a refresh there writes short-lived STS credentials into the caller's own secret directory, which is what #195 reports. It is single-user rather than shared, and `secret_refresh_persistent.test` currently asserts the refreshed material survives a restart. Moving it would close #195 as well, but it is a deliberate change to a tested guarantee — so I have raised it rather than decided it in the diff. Happy to extend either way.

**In-place would be better.** #412 suggests updating the loaded secret's contents in place, which would last the session rather than one transaction, so a long-lived session would re-run the credential chain once instead of once per transaction. That needs a new `SecretManager` API in core — the existing `// TODO: change SecretManager API to avoid requiring catching this exception` points at the same seam. This fix works within the current API and does not block that.

## Tests

New `s3_refresh_storage_test.cpp` registers a stand-in extension storage (persistent, shared, optionally read-only) via `SecretManager::LoadSecretStorage` and seeds a refreshable secret with `CREATE SECRET … IN <storage>`:

* the refresh writes nothing to the store, the store's copy keeps the stale material, the refreshed copy lands in `transaction` storage and wins the lookup, `GetSecretByName` raises no ambiguity, and after `COMMIT` only the untouched store copy remains;
* the same against a read-only store, which previously threw;
* end-to-end through the mock S3 server, writable and read-only: the 403 is retried with the fresh key and the read succeeds.

All three fail without the change.

---

### DIFF
--- a/src/create_secret_functions.cpp
+++ b/src/create_secret_functions.cpp
@@ -4,6 +4,7 @@
 #include "s3/s3fs.hpp"
 #include "duckdb/main/extension/extension_loader.hpp"
 #include "duckdb/common/local_file_system.hpp"
+#include "duckdb/main/secret/secret_manager.hpp"
 
 namespace duckdb {
 
@@ -109,6 +110,16 @@ unique_ptr<BaseSecret> CreateS3SecretFunctions::CreateSecretFunctionInternal(Cli
 	return S3SecretBuilder(input).Create();
 }
 
+//! Only the secret manager's built-in storages. Re-creating a refreshed secret into an extension-registered storage
+//! writes session-local credentials to a shared store: it throws where the session cannot write, and silently mutates
+//! the store where it can. See duckdb/duckdb-httpfs#412.
+static bool RefreshMayWriteToStorage(const string &storage_mode) {
+	auto storage = Identifier(storage_mode);
+	return storage.empty() || storage == SecretManager::TEMPORARY_STORAGE_NAME ||
+	       storage == SecretManager::LOCAL_FILE_STORAGE_NAME || storage == SecretManager::TRANSACTION_STORAGE_NAME ||
+	       storage == SecretManager::CONNECTION_STORAGE_NAME;
+}
+
 CreateSecretInput CreateS3SecretFunctions::GenerateRefreshSecretInfo(const SecretEntry &secret_entry,
                                                                      Value &refresh_info) {
 	const auto &kv_secret = secret_entry.secret->Cast<KeyValueSecret>();
@@ -120,7 +131,10 @@ CreateSecretInput CreateS3SecretFunctions::GenerateRefreshSecretInfo(const Secre
 	result.type = kv_secret.GetType();
 	result.name = kv_secret.GetName();
 	result.provider = Identifier(kv_secret.GetProvider());
-	if (result.persist_type != SecretPersistType::TRANSACTION) {
+	if (result.persist_type == SecretPersistType::TRANSACTION || !RefreshMayWriteToStorage(secret_entry.storage_mode)) {
+		// No storage write; storage_type must stay empty, TRANSACTION secrets reject an explicit storage.
+		result.persist_type = SecretPersistType::TRANSACTION;
+	} else {
 		result.storage_type = Identifier(secret_entry.storage_mode);
 	}
 	result.scope = kv_secret.GetScope();
--- a/test/unittest/s3/CMakeLists.txt
+++ b/test/unittest/s3/CMakeLists.txt
@@ -1,6 +1,7 @@
 target_sources(
   unittest_httpfs
   PRIVATE mock_s3_server.cpp
+          s3_refresh_storage_test.cpp
           s3_refresh_test.cpp
           s3_request_test.cpp
           s3_retry_test.cpp
--- a/test/unittest/s3/s3_refresh_storage_test.cpp
+++ b/test/unittest/s3/s3_refresh_storage_test.cpp
@@ -1,0 +1,312 @@
+#include "catch.hpp"
+
+#include "s3/mock_s3_server.hpp"
+#include "s3/s3_test_helper.hpp"
+
+#include "create_secret_functions.hpp"
+
+#include "duckdb.hpp"
+#include "duckdb/catalog/catalog_transaction.hpp"
+#include "duckdb/common/file_system.hpp"
+#include "duckdb/common/string_util.hpp"
+#include "duckdb/main/client_context.hpp"
+#include "duckdb/main/secret/secret_manager.hpp"
+#include "duckdb/main/secret/secret_storage.hpp"
+
+namespace duckdb {
+
+namespace {
+
+constexpr const char *SHARED_STORE_NAME = "test_shared_store";
+constexpr const char *SHARED_SECRET_NAME = "shared_store_s3";
+
+//! Tie-break offset for the stand-in store. Must not collide with the built-ins (transaction 0, connection 5,
+//! memory 10, local_file 20) -- the secret manager rejects a storage whose offset already exists.
+constexpr int64_t SHARED_STORE_OFFSET = 30;
+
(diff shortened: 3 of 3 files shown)


kinghuang:
Opened #416 as an alternative to this PR, so the two approaches can be compared directly. Both fix the correctness bug in #412 completely — no throw on a read-only store, no write to a shared store. The difference is what happens afterwards.

This PR lands the refreshed copy in the transaction storage, which is the only shadowing storage available without a core change. That copy is dropped at each commit, so a long-lived session re-pays the failed-auth round-trip and re-runs the credential chain in every transaction. Measured against the mock S3 server, three reads in three transactions:

| secret storage | stale 403s | credential-chain resolutions |
|---|---|---|
| `memory` (unchanged by this PR) | 4 | 1 |
| extension store, this PR | **12** | **3** |
| extension store, #416 | 4 | 1 |

`HTTPState::RunCredentialRefresh` is only a mutex, so nothing absorbs that.

#416 depends on duckdb/duckdb#25191, which adds a session-scoped `refresh` secret storage that shadows the original — the in-place update #412 asks for, expressed as an overlay rather than a mutation so it works for storages the secret manager does not own the cache for. With it the rule collapses to "a refresh never writes to a secret storage", the per-storage carve-out list here disappears, and #195 is resolved as well.

The trade is a core change and a submodule bump versus a self-contained fix that ships today. Happy either way — this PR stands on its own if the core surface is not wanted, and #416 is a draft until duckdb/duckdb#25191 lands.

The `local_file` question I raised above is answered by #416 and left open here.

kinghuang:
Closing in favour of #416.

@lnkuiper [prefers that direction](https://github.com/duckdb/duckdb-httpfs/pull/416#issuecomment-5568030872), and neither PR lands before the core refresh rework he describes there. This one routes the refresh to the transaction storage precisely because it was the only shadowing storage available without a core change — which makes it an instance of the workaround that rework is meant to remove, so there is no reason to keep it open as a fallback.

The diagnosis, the reproduction and the measurements stay in #412.

PR #416, opened 2026-08-31
Alternative to #413: refresh S3 secrets into a session-scoped overlay

### LINKED ISSUE #195: Using STS assume role with persistent secret fails

kinghuang:
**Alternative to #413 — pick one.** Same bug (#412), different mechanism. This one also resolves #195.

> ⚠️ **Depends on duckdb/duckdb#25191 and does not build against the pinned duckdb.** CI here will fail to compile (`REFRESH_STORAGE_NAME` is not yet a member of `SecretManager`) until that core PR lands and the submodule pin moves — the pin and `duckdb_version` are deliberately left untouched so this diff shows only the httpfs delta. Ready for review as a direction to choose, not as a branch to merge today.

## The difference in one line

| | #413 | this |
|---|---|---|
| needs a core change | no | **yes** (duckdb/duckdb#25191) |
| refresh writes to a storage | never, for extension storages | **never, for any storage** |
| refreshed copy lives for | one transaction | **the session** |
| resolves #195 (`local_file`) | no | **yes** |
| stale round-trips over 3 reads in 3 transactions | 12 | **4** |

Both fix the correctness bug in #412 completely. The difference is that #413's copy is dropped at each commit, so a long-lived session re-pays the failed-auth round-trip and re-runs the credential chain in every transaction. Measured against the mock S3 server, three reads in three transactions: 12 stale 403s and three credential-chain resolutions under #413, versus 4 and one here — the same cost as a `memory`-backed secret. `HTTPState::RunCredentialRefresh` is only a mutex, so nothing absorbs that.

## The change

duckdb/duckdb#25191 adds a `refresh` secret storage: session-scoped, written by no SQL, and *shadowing* — by-name resolution reaches it ahead of the original instead of reporting the two copies as ambiguous, and a `DROP SECRET` takes the shadow with it. That is the in-place update #412 asks for, expressed as an overlay rather than a mutation, so it works uniformly for every storage including extension-registered ones that the secret manager does not own the cache for.

`GenerateRefreshSecretInfo` then collapses to one rule — **a refresh never writes to a secret storage**:

```cpp
if (result.persist_type != SecretPersistType::TRANSACTION) {
	result.persist_type = SecretPersistType::TEMPORARY;
	result.storage_type = Identifier(SecretManager::REFRESH_STORAGE_NAME);
}
```

No `RefreshMayWriteToStorage` list, no per-storage carve-outs. #413 needs that list because transaction storage is the only shadowing storage available today, and routing `local_file` there would have been a behaviour change with nothing to show for it.

## What this changes that #413 does not

`test/sql/secret/secret_refresh_persistent.test` is updated: a persistent secret's refresh no longer lands on disk. Both copies stay visible, so the shadowing is inspectable:

```
┌──────┬────────────┬────────────┬──────────────────┐
│ name │ persistent │  storage   │ has fresh key_id │
├──────┼────────────┼────────────┼──────────────────┤
│ s1   │ true       │ local_file │ false            │
│ s1   │ false      │ refresh    │ true             │
└──────┴────────────┴────────────┴──────────────────┘
```

After a restart only the `local_file` row remains, still carrying the original material. That is the half of #195 where the same write fails outright for an `sts`-chain persistent secret.

## Tests

Everything in #413, plus two that only pass here:

* **`refreshes a shared-storage S3 secret once per session`** — three reads in three transactions drive exactly one credential-chain resolution and one set of stale round-trips, pinning the number above.
* **`refreshes a persistent S3 secret without writing to its storage`** — the `local_file` copy still holds the original material while the overlay carries the refreshed one.

The shared-storage test also now asserts the refresh survives `COMMIT`, and that `DROP SECRET` removes the overlay copy so no stale shadow outlives the secret.

httpfs unit suite: 76 cases, 7031 assertions, green. Core suites for duckdb/duckdb#25191 are listed on that PR.

---

### DIFF
--- a/src/create_secret_functions.cpp
+++ b/src/create_secret_functions.cpp
@@ -4,6 +4,7 @@
 #include "s3/s3fs.hpp"
 #include "duckdb/main/extension/extension_loader.hpp"
 #include "duckdb/common/local_file_system.hpp"
+#include "duckdb/main/secret/secret_manager.hpp"
 
 namespace duckdb {
 
@@ -121,7 +122,10 @@ CreateSecretInput CreateS3SecretFunctions::GenerateRefreshSecretInfo(const Secre
 	result.name = kv_secret.GetName();
 	result.provider = Identifier(kv_secret.GetProvider());
 	if (result.persist_type != SecretPersistType::TRANSACTION) {
-		result.storage_type = Identifier(secret_entry.storage_mode);
+		// A refresh re-resolves session-local credentials, so it never writes to the storage the secret came from:
+		// the refreshed copy lands in the session-scoped refresh overlay and shadows the original.
+		result.persist_type = SecretPersistType::TEMPORARY;
+		result.storage_type = Identifier(SecretManager::REFRESH_STORAGE_NAME);
 	}
 	result.scope = kv_secret.GetScope();
 
--- a/test/sql/secret/secret_refresh_persistent.test
+++ b/test/sql/secret/secret_refresh_persistent.test
@@ -62,26 +62,25 @@ FROM "s3://test-bucket/persistent-secret-refresh.parquet"
 ----
 42
 
+# The refresh lands in the session-scoped overlay and shadows the stored secret; the persistent copy is untouched.
 query IIII
 SELECT name, persistent, storage, secret_string LIKE '%key_id={AWS_ACCESS_KEY_ID}%'
 FROM duckdb_secrets(redact=false)
 WHERE name='s1'
+ORDER BY storage
 ----
-s1	true	local_file	true
-
-query I
-SELECT count(*) FROM duckdb_secrets(redact=false) WHERE name='s1'
-----
-1
+s1	true	local_file	false
+s1	false	refresh	true
 
 restart
 
 statement ok
 SET secret_directory='{TEST_DIR}/secret_refresh_persistent'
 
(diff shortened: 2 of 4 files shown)


lnkuiper:
Thanks for the PR! I prefer this over #413, but the DuckDB core PR is not in the shape we want. We want to implement a proper refresh mechanism, instead of the somewhat hacky system we use now. We will send a proper structural fix in DuckDB core before we can fix this.

kinghuang:
Thanks @lnkuiper, that makes sense. I've closed both #413 and duckdb/duckdb#25191. I'm happy to help with changes on the duckdb-httpfs side once the structural changes in DuckDB core land.

PR #434, opened 2026-09-10
Bump remove version

carlopi:


carlopi:
result: `duckdb_version: ''` works, but it feels a bit hacky. Unsure what's the proper path here.

PR #442, opened 2026-09-17
Parallelize S3 glob listing

pdet:
S3 globbing lists directory prefixes sequentially, which can delay queries over datasets spread across many directories.

This PR uses the async scheduler to prefetch prefix pages concurrently. Prefetching stays bounded and preserves lazy listing. Files are returned in key order across pagination and nested directories, regardless of async_threads.


carlopi:
Looks super cool / interesting.

Main question is whether we can get into trouble with S3 listing 503 / how to throttle / recover there. That was likely also needed before, but now we are more hungry.

carlopi (approved):
Thanks, in particular https://github.com/duckdb/duckdb-httpfs/pull/442/commits/3773b02b15e9f794297951faf51f3414ce00911c (and the plumbing in duckdb/duckdb) make a lot of sense to me, to avoid this being too aggressive on List limits.

PR #443, opened 2026-09-17
Disable curl signal handling for concurrent requests

pdet:
This PR sets CURLOPT_NOSIGNAL when each curl handle is created, preventing those signal-handler changes. The setting remains active when handles are reused.


carlopi:
Thanks!

PR #445, opened 2026-09-18
Port #399 (cache certificate store) onto current main

carlopi:
Original at #399, from @smvv.


carlopi:
To be reviewed: interactions on MacOS with `NATIVE_CA`

carlopi:
I think proper solution is:
1. patch to CURL (via vcpkg)
2. out out to `CURLSSLOPT_NATIVE_CA`

I'll fix this later.

